PR #5, opened 2024-09-12
Replace asset check with amount check for asset transfer deep links

mitsinsar:


versions.gradle
```diff
@@ -60,6 +60,6 @@ ext {
 
     peraApplication = [
             versionCode: 118,
-            versionName: "5.10.13"
+            versionName: "90.20.772-1"
```
ehanoc: What does this number represent?

app/src/main/java/com/algorand/android/modules/deeplink/domain/model/BaseDeepLink.kt
```diff
@@ -196,7 +196,7 @@ sealed class BaseDeepLink {
 
             override fun doesDeeplinkMeetTheRequirements(rawDeepLink: RawDeepLink): Boolean {
                 return with(rawDeepLink) {
-                    val doesDeeplinkHaveAssetTransferQueries = accountAddress != null && assetId != null
+                    val doesDeeplinkHaveAssetTransferQueries = accountAddress != null && amount != null
```
ehanoc: Not need for asset check? We assume it's always algo?

versions.gradle
mitsinsar: 90 means internal release
20 is sprint
772 is ticket number
- 1 is the iteration of the ticket.

app/src/main/java/com/algorand/android/modules/deeplink/domain/model/BaseDeepLink.kt
mitsinsar: Checking amount is enough to see that this is an asset transfer deep link. And while creating the asset transfer deep link object, we get asset id `if exists`,  if not, we say this is an Algo transfer deep link

versions.gradle
ehanoc: Okay, why is it being updated here? versioning should only be updated for releases and as part of the building step ideally. no?

app/src/main/java/com/algorand/android/modules/deeplink/domain/model/BaseDeepLink.kt
ehanoc: > while creating the asset transfer deep link object

While creating, not verifying? Sorry, not sure i understood how we check if it's a simple algo transfer or an asset transfer here.

versions.gradle
mitsinsar: We have two CI/CD trigger patterns;
`*.*.*-*` -> Triggers internal release workflow
`*.*.*` -> Triggers store release workflow

This fix will be a hot fix, that's why I triggered internal test release immediately.

app/src/main/java/com/algorand/android/modules/deeplink/domain/model/BaseDeepLink.kt
mitsinsar: In this step, we don't need to check if this is an asset transfer or Algo transfer. We don't have two different asset transfer models like `AssetTransferDeepLink` and `AlgoTransferDeepLink`, we just have `AssetTransferDeepLink`. That's why it is enough to check if this is a `transfer` (has account address and amount) deep link or not. 

And while creating `AssetTransferDeepLink` object, we simply check if deep link has `asset` query or not. If there is, we parse it, if not, we say this is an Algo transfer and pass Algo asset id.

Something like;
```
AssetTransferDeepLink(
    assetId = if (rawDeeplink.assetId) != null rawDeeplink.assetId else ALGO_ASSET_ID,
    ...
)
```

yasin-ce (approved):

PR #6, opened 2024-09-18
Fix banner color issue on dark mode

yasin-ce:
Update banner text colors according to the design file.

### DIFF
--- a/versions.gradle
+++ b/versions.gradle
@@ -60,6 +60,6 @@ ext {
 
     peraApplication = [
             versionCode: 118,
-            versionName: "90.19.764-2"
+            versionName: "90.19.764-3"
     ]
 }
--- a/versions.gradle
+++ b/versions.gradle
@@ -60,6 +60,6 @@ ext {
 
     peraApplication = [
             versionCode: 118,
-            versionName: "90.19.764-4"
+            versionName: "90.19.764-5"
     ]
 }
--- a/versions.gradle
+++ b/versions.gradle
@@ -60,6 +60,6 @@ ext {
 
     peraApplication = [
             versionCode: 118,
-            versionName: "90.19.764-1"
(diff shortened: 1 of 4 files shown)


ehanoc (approved):

versions.gradle
```diff
@@ -60,6 +60,6 @@ ext {
 
     peraApplication = [
             versionCode: 118,
-            versionName: "5.10.13"
+            versionName: "90.19.764-5"
```
ehanoc: To avoid conflicts on every PR; we can just update this on releases? Unnecessary for `dev` branch

ehanoc (changes requested):

versions.gradle
yasin-ce: We are sending test releases for each ticket. That's why we are updating this value on each ticket branch. Is there any other alternative solution for this?

michaeltchuang: > To avoid conflicts on every PR; we can just update this on releases? Unnecessary for `dev` branch

@yasin-ce - I also agree with @ehanoc on this

Maybe this change should be in another PR, but don't manually change the app version name in code, this should be scripted into the CI job so it can be automated for any feature branch builds.  I suggest using the PR branch name and github action run number to help generate the version name (e.g. you can pull this info from github workflow environment variables)

in build.gradle
```
android {
  versionName buildVersionName()
}

def buildVersionName() {
   return "computed version name from branch name"
}
```

Also, the internal version name does seem verbose too (why does sprint number matter?  what happens if a ticket rollovers to next sprint if not merged...do you keep incrementing version name?)

yasin-ce: Thanks for the recommendations. I'll create a separate task for this. I will do research and improve the flows. 
@michaeltchuang @ehanoc

ehanoc (approved):

PR #55, opened 2024-12-06
PERA-977 :: Add ASA ID next to shortName in description

michaeltchuang:
# Pull Request Template

## Description
- Add ASA ID next to shortName in description

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-977

## Checklist

## Additional Notes
- When i used the `*`, formatting made it move couple pixels up so i just used `-` instead.

### DIFF
--- a/app/src/main/java/com/algorand/android/modules/accountdetail/assets/ui/adapter/OwnedAssetViewHolder.kt
+++ b/app/src/main/java/com/algorand/android/modules/accountdetail/assets/ui/adapter/OwnedAssetViewHolder.kt
@@ -35,7 +35,9 @@ class OwnedAssetViewHolder(
                     )
                 }
                 setTitleText(name.getName(resources))
-                setDescriptionText(shortName.getName(resources))
+                val name = shortName.getName(resources).uppercase()
+                val descriptionText = if (name == "ALGO") name else "$name  -  ${item.id}"
+                setDescriptionText(descriptionText)
                 setPrimaryValueText(formattedAmount)
                 setSecondaryValueText(if (isAmountInDisplayedCurrencyVisible) formattedDisplayedCurrencyValue else null)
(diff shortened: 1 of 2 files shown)


michaeltchuang:
Here is screenshot of it

Screenshot_1733457695

app/src/main/java/com/algorand/android/modules/accountdetail/assets/ui/adapter/OwnedAssetViewHolder.kt
```diff
@@ -35,7 +35,9 @@ class OwnedAssetViewHolder(
                     )
                 }
                 setTitleText(name.getName(resources))
-                setDescriptionText(shortName.getName(resources))
+                val name = shortName.getName(resources).uppercase()
+                val descriptionText = if (name == "ALGO") name else "$name  -  ${item.id}"
```
mitsinsar: If -somehow- we change asset naming rule, that would break this implementation

Instead of name check, it would be better if we can add a boolean in `BaseOwnedItem.AssetItem`. 
```
val isAlgo: Boolean
    get() = id == ALGO_ID
```

mitsinsar (changes requested):

app/src/main/java/com/algorand/android/modules/accountdetail/assets/ui/adapter/OwnedAssetViewHolder.kt
michaeltchuang: good suggestion, i pushed up change

mitsinsar:
As far as I know, unit name limit is 8 bytes, so we don't need to think about "long" unit names. But if you think there is a way to create/display long "unit name", asset id will be ellipsized.

mitsinsar (approved):

PR #92, opened 2024-12-28
PERA-1348 :: Add in architecture related ER diagrams designed in mermaid

michaeltchuang:
# Pull Request Template

## Description
- Add in architecture ER diagrams designed in mermaid tool

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1348

## Checklist

## Additional Notes
- Any modifications to diagram syntax should be viewable in mermaid


ARCHITECTURE.md
```diff
@@ -0,0 +1,65 @@
...
+    no_auth {
+        String encrypted_address PK
+    }
+    hd_seeds {
+        Int seed_id PK
+        ByteArray encrypted_seed UK
+        String seed_custom_name
+    }
+    hd_keys {
+        String encrypted_address PK
```
ehanoc, line 58: @michaeltchuang @mitsinsar Lets not use address, but public keys

ehanoc: Lets also avoid strings and any encoding to store raw public keys. Bytes :)

michaeltchuang: > @michaeltchuang @mitsinsar Lets not use address, but public keys

@ehanoc I actually had `public_key` as PK at first for `hd_keys` table...but I realized when we fetch/delete through Pera UI or deeplink, user may want to do this by algorand address...so I think algorand address should be a column in table.  Is it easy to go `algorand address -> public key`?   If so, then maybe I can create a mapper back/forth.

@yigitguler are you back this week?  Did you want to weigh in what you think should be stored and what should be derived runtime?

michaeltchuang: I updated the PR now so hd code/schema is using public keys and algorand addresses are derived runtime

yigitguler: @michaeltchuang Algorand address can be seen as a denormalized field in the table. IMO, it is ok to keep the Algorand address in the table as well, only if it brings significant performance optimization.

yigitguler: In a scenario where the user has 100 accounts (I know it is above our current limits) How much computation time does it require to get the algorand address from the public key? If it is really fast, maybe we can just store the public key and prepare some utility functions to lookup the account with the algorand address.

mitsinsar (approved):

ARCHITECTURE.md
```diff
@@ -0,0 +1,152 @@
...
+    no_auth {
+        String encrypted_address PK
+    }
+    hd_entropy {
+        Int entropy_id PK
+        String encrypted_entropy UK
+        String entropy_custom_name
+    }
+    hd_addresses {
+        String encrypted_address PK
```
ehanoc, line 58: @michaeltchuang Why encrypted address & public key? there's zero risk and adds extra steps at no gain.

PR #112, opened 2025-01-21
PERA-1226 :: Move QR code scanner and feature toggle into common module

mitsinsar:
## Summary
- Move QR code scanner and feature toggle into common module


michaeltchuang (approved):
i'll merge this PR first so tests can pass in parent PR.  We can fix things in future in parent PR

PR #124, opened 2025-02-02
PERA-1550 :: Migrate algo sdk into common-sdk

mitsinsar:
# Pull Request Template

## Description
- Migrated necessary classes and function into common-sdk

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1550
- Due to go library, couldn't test go lib related implementations, that's why I hid them behind an AlgoSdk interface to minimize untested files.

## Checklist
- [ ] Have you documented any necessary changes in the project's documentation?

## Additional Notes
- None of the implementations will be used by production code for now. Only used reference is algo sdk account validation


michaeltchuang (changes requested):
can we avoid creating a new module for an `aar` file?  let's just add a libs folder in common-sdk and put `aar` file there (until we get a mavencentral version)

also maybe use api in common-sdk then we don't need to add a reference in `app/build.gradle`

mitsinsar:
Wanted to extract into another module, so that we could share it with other modules in the future, but it is not necessary right now. Updated as you mentioned 👍 

Oh! I also remembered another reason;

```
   > Direct local .aar file dependencies are not supported when building an AAR. 
The resulting AAR would be broken because the classes and Android resources from any 
local .aar file dependencies would not be packaged in the resulting AAR. Previous versions
 of the Android Gradle Plugin produce broken AAR  in this case too (despite not throwing this error). 
```

michaeltchuang (approved):
i see what you meant by lint not liking aar files in library module

PR #115, opened 2025-01-23
PERA-1506 :: Pass custom fee for keyreg online transactions

michaeltchuang:
# Pull Request Template

## Description
- Pass custom fee for keyreg online transactions

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1506

## Checklist
- [X] Have you tested your changes locally?
- [X] Have you reviewed the code for any potential issues?
- [X] Have you documented any necessary changes in the project's documentation?
- [X] Have you added any necessary tests for your changes?
- [X] Have you updated any relevant dependencies?

## Additional Notes
- Add any additional notes or comments that may be helpful for reviewers.

PR #141, opened 2025-02-17
PERA-1353 :: Add new bottomsheet for HD Wallet flow

### LINKED ISSUE #50: Task - New Account Recovery UI Flow
For HD Wallet feature, we have a new account recovery flow.  Please refactor current screen done in xml and design new one using compose.  Implement code first in `app` module, but once you get it working...try to put as much code into common `wallet-sdk` module when creating PR since we want this to be reusable in other apps

1. Refactor first screen from xml to compose
2. Design new bottom sheet that allows user to pick BIP39 or Algo25 
3. Refactor third screen from xml to compose (should handle 24 or 25 words)

image

### LINKED ISSUE #142: Task - Convert RegisterIntroFragment to Compose
For HD Wallet feature, we have a new create wallet flow.  Please refactor this screen coded in xml and convert it to compose so we can use our new bottom sheet designed in recover flow (https://github.com/perawallet/pera-android/issues/50).

Here is the screenshot of the xml screen

Image

We will use it like this in future

Image

michaeltchuang:
# Pull Request Template

## Description
- Add new bottomsheet for future create and recovery HD Wallet flow
- The bottomsheet is hidden by feature flag (only enabled for non prod release builds)
- QR code recovery will input 24 words (recover button afterwards doesn't work or proceed)

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1353
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1666
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1707

## Checklist

## Additional Notes
- Compose widget/theme code should be moved out of app module to it's own module eventually
- We can polish colors if needed later, but I think this is usable for now

Here are some screenshots of bottom sheet

Screenshot_1739833474
Screenshot_1739833532


michaeltchuang:
for create wallet, we'll need to show recommended label and here are some images of that

Screenshot_1739838252
Screenshot_1739838277

app/src/test/kotlin/com/algorand/android/modules/algosdk/domain/mapper/TransactionParametersResponseMapperImplTest.kt
```diff
@@ -1,43 +0,0 @@
...
-
-package com.algorand.android.modules.algosdk.domain.mapper
-
-import com.algorand.algosdk.util.Encoder
-import com.algorand.android.models.TransactionParams
-import kotlin.test.assertEquals
-import kotlin.test.assertTrue
-import org.junit.Test
-
-class TransactionParametersResponseMapperImplTest {
```
mitsinsar: Even tho this class is not used in dev branch, `TransactionParametersResponseMapperImpl` exists in epic branch. Are we sure about deleting this class?

app/src/main/res/values/strings.xml
```diff
@@ -264,12 +264,12 @@
     <string name="im_ready_to_begin">I\'m Ready to Begin</string>
     <string name="create_a_watch">Create a Watch Account</string>
     <string name="recover_an_algorand">Recover an Algorand account</string>
-    <string name="in_the_following">In the following steps, you\'ll enter your 25-word recovery passphrase to recover an Algorand account. Have your recovery phrase handy before you begin.</string>
+    <string name="in_the_following">In the following steps, you\'ll enter your recovery passphrase to recover an Algorand account. Have your recovery phrase handy before you begin.</string>
```
mitsinsar, line 277: We are updating this but is this approved? this change will effect prod as well, as well as the one below

app/src/main/kotlin/com/algorand/android/ui/compose/widget/MnemonicTypeCard.kt
```diff
@@ -0,0 +1,116 @@
...
+                            .padding(start = 10.dp)
+                            .clip(shape = CircleShape)
+                            .background(color = MaterialTheme.colorScheme.surfaceVariant)
+                    ) {
+                        PeraBodyText(
+                            modifier = Modifier.padding(
+                                start = 7.dp,
+                                end = 7.dp,
+                                top = 3.dp,
+                                bottom = 3.dp
```
mitsinsar: This can be `vertical = 7.dp ` and `horizontal 3.dp`

app/src/main/kotlin/com/algorand/android/ui/compose/widget/MnemonicTypeCard.kt
```diff
@@ -0,0 +1,116 @@
...
+                            ),
+                            text = it
+                        )
+                    }
+                }
+            }
+
+            Row(
+                modifier = Modifier.padding(top = 16.dp, bottom = 16.dp),
```
mitsinsar: This can be `vertical = 16.dp`

app/src/main/kotlin/com/algorand/android/ui/compose/theme/PeraTheme.kt
```diff
@@ -42,7 +42,7 @@ private val lightScheme = lightColorScheme(
     surfaceContainerLow = PeraColor.surfaceContainerLowLight,
     surfaceContainer = PeraColor.surfaceContainerLight,
     surfaceContainerHigh = PeraColor.surfaceContainerHighLight,
-    surfaceContainerHighest = PeraColor.surfaceContainerHighestLight
+    surfaceContainerHighest = PeraColor.surfaceContainerHighestLight,
```
mitsinsar: Unnecessary comma

app/src/main/kotlin/com/algorand/android/ui/compose/theme/PeraTheme.kt
```diff
@@ -80,7 +80,7 @@ private val darkScheme = darkColorScheme(
     surfaceContainerLow = PeraColor.surfaceContainerLowDark,
     surfaceContainer = PeraColor.surfaceContainerDark,
     surfaceContainerHigh = PeraColor.surfaceContainerHighDark,
-    surfaceContainerHighest = PeraColor.surfaceContainerHighestDark
+    surfaceContainerHighest = PeraColor.surfaceContainerHighestDark,
```
mitsinsar: Unnecessary comma

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
```diff
@@ -45,27 +75,137 @@ class RecoverAccountInfoFragment : BaseInfoFragment() {
...
+            androidx.compose.material3.ModalBottomSheet(
+                onDismissRequest = { showBottomSheet.value = false },
+                sheetState = sheetState,
+                containerColor = MaterialTheme.colorScheme.surface,
+                contentColor = MaterialTheme.colorScheme.onSurface
+            ) {
+                BottomSheetContent(sheetState)
+            }
+        }
+    }
```
mitsinsar: This says `PrimaryButton` however it also handles bottom sheet logic. I think we shouldn't have any composable as `override fun` at all. But if we do and it says `primary button` then it should provide `primary button` only.

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
```diff
@@ -45,27 +75,137 @@ class RecoverAccountInfoFragment : BaseInfoFragment() {
...
+                Icon(
+                    imageVector = Icons.Filled.Close,
+                    tint = MaterialTheme.colorScheme.primary,
+                    contentDescription = "Close"
+                )
+            }
+            Spacer(Modifier.weight(0.1f))
+
+            PeraTitleText(
+                text = "Select your Mnemonic type"
```
mitsinsar: We should get this from string resource

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
```diff
@@ -45,27 +75,137 @@ class RecoverAccountInfoFragment : BaseInfoFragment() {
...
+                .clip(RoundedCornerShape(16.dp))
+                .background(MaterialTheme.colorScheme.surface)
+                .padding(16.dp)
+        ) {
+            BottomSheetHeader(sheetState)
+
+            MnemonicTypeCard(
+                title = "Bip39",
+                description = "New inter-operable format that enables important features like HD Wallet",
+                footer = "24 Key mnemonic keys",
```
mitsinsar: We should get these strings from string resource

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
```diff
@@ -45,27 +75,137 @@ class RecoverAccountInfoFragment : BaseInfoFragment() {
...
+                    coroutineScope.launch {
+                        sheetState.hide()
+                    }
+                }
+            )
+
+            MnemonicTypeCard(
+                title = "Algo25",
+                description = "Legacy format that is specific to Algorand ecosystem",
+                footer = "25 Key mnemonic keys",
```
mitsinsar: We should get these strings from string resource

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoveryAccountInfoViewModel.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@HiltViewModel
+class RecoveryAccountInfoViewModel @Inject constructor(
+    private val getIsProductionReleaseUseCase: GetIsProductionReleaseUseCase,
+    private val getIsActiveNodeTestnetUseCase: GetIsActiveNodeTestnetUseCase,
+    private val isFeatureToggleEnabled: IsFeatureToggleEnabled
+) : ViewModel() {
+
+    fun isHdWalletToggleEnabled(): Boolean {
+        val isHdWalletToggleEnabled = isFeatureToggleEnabled(HD_WALLET_BUTTON_TOGGLE) &&
```
mitsinsar: I think we will have this flag in multiple places. We can create a usecase and move this logic into use case to prevent from having duplications

app/src/main/res/navigation/login_navigation.xml
```diff
@@ -111,6 +111,11 @@
                 android:defaultValue="@null"
                 app:argType="string"
                 app:nullable="true" />
+            <argument
```
mitsinsar, line 117: This implementation creates potential bugs. There is already mnemonic args. We can pass 24 mnemonic and 25 word counts.

Arg type can be a sealed interface with 2 different classes that implement this interface. passing two different arg that serves to same purpose.

app/src/main/res/navigation/recover_with_passphrase_navigation.xml
```diff
@@ -28,6 +28,11 @@
             android:defaultValue="@null"
             app:argType="string"
             app:nullable="true" />
+        <argument
```
mitsinsar, line 31: This implementation creates potential bugs. There is already mnemonic args. We can pass 24 mnemonic and 25 word counts.

Arg type can be a sealed interface with 2 different classes that implement this interface. passing two different arg that serves to same purpose.

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/usecase/RecoverWithPassphrasePreviewUseCase.kt
```diff
@@ -61,8 +64,11 @@ class RecoverWithPassphrasePreviewUseCase @Inject constructor(
         clipboardData: String
     ): RecoverWithPassphrasePreview {
         val splittedText = clipboardData.splitMnemonic()
-        return if (splittedText.size != ACCOUNT_PASSPHRASES_WORD_COUNT) {
-            preview.copy(onGlobalErrorEvent = Event(R.string.the_last_copied_text))
+        return if (
+            splittedText.size != ALGO25_WALLET_PASSPHRASES_WORD_COUNT &&
+            splittedText.size != HD_WALLET_PASSPHRASES_WORD_COUNT
+            ) {
```
mitsinsar: This implementation doesn't seem right. What if there is another type of account that uses. `24` or `25` words? We should send wallet type instead of integer that represent wallet

mitsinsar (changes requested):

app/src/main/kotlin/com/algorand/android/ui/compose/theme/PeraTheme.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/ui/compose/theme/PeraTheme.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/ui/compose/widget/MnemonicTypeCard.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/ui/compose/widget/MnemonicTypeCard.kt
michaeltchuang: you got it backwards, but fixed

app/src/main/res/values/strings.xml
michaeltchuang: this is how string is in figma file, I fixed the 2nd string so it can go out to prod earlier without issue

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoverAccountInfoFragment.kt
michaeltchuang: we're currently building off `BaseInfoFragment` since that's the existing style...I'm not sure we should move away from overriding till we have more bandwidth.  I'm open minded though, maybe we can talk tomorrow in team meeting about this.   

Basically 12+ references are just copy changes on top of `BaseInfoFragment` and maybe two screens will use bottomsheet for now.

app/src/test/kotlin/com/algorand/android/modules/algosdk/domain/mapper/TransactionParametersResponseMapperImplTest.kt
michaeltchuang: yeah, looks like we can safely delete it.  That file was being used by KeyReg stuff, but when we moved to Go SDK for that...Yasin removed it

https://github.com/perawallet/pera-android/pull/115/files#diff-b2462e664b5c26e255327a2f3f60b01651d8b3867256cb393ffd01133ff1bddd

let me update the epic branch tomorrow...it's was worth double-checking

app/src/main/res/navigation/recover_with_passphrase_navigation.xml
michaeltchuang: one variable is the input mnenomic string (empty for user to enter into form or qr code value that prefills the form)...the other variable is number of word text fields to show on the passphrase form.  So instead of passing two input variables you want one?  We can chat about this tomorrow in team meeting

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/info/ui/RecoveryAccountInfoViewModel.kt
michaeltchuang: I switched to use case now

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/usecase/RecoverWithPassphrasePreviewUseCase.kt
michaeltchuang: i switched to using `AccountType` and added the world count value there

app/src/main/res/navigation/recover_with_passphrase_navigation.xml
michaeltchuang: I switched to passing `AccountType` instead of word count so i think this issue is addressed

app/src/main/res/navigation/login_navigation.xml
michaeltchuang: I switched to passing `AccountType` instead of word count so i think this issue is addressed.  QR code parsing will default to word count when split checking since we don't know what account type it is (24 or 25 word passphrase)

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
```diff
@@ -0,0 +1,107 @@
...
+    @get:Rule
+    var hiltRule = HiltAndroidRule(this)
+
+    @Inject
+    lateinit var aesPlatformManager: AESPlatformManager
+
+    @Before
+    fun setup() {
+        hiltRule.inject()
+    }
```
mitsinsar: With this, we are testing provided AESPlatformManager. What we need to do is to create instance of the object manually like;

```
private val sut = AESPlatformManagerImpl(...)
```

1. This adds extra processing which makes test slower
2. This tests the implementation of the real class. Unit test needs to be isolated, that's why we should be testing implementation, not provided interface

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    @Test
+    fun testEncryptDecryptAlgo25SecretKey_shouldMatchOriginal() {
+        // Given
+        val originalSecretKey = Sdk.generateSK()
```
mitsinsar: We shouldn't use `Sdk` functions, or any other library functions in our unit tests. By calling this function, 
- If `Sdk.generateSK` fails, the test will fail as well, even if `sut` works perfectly.
- Any change in `Sdk.generateSK` may break out tests. 

Our tests should be independent of external libraries

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        // Debugging: Verify secret key equality
+        assertTrue(originalSecretKey.contentEquals(decryptedSecretKey))
+    }
+
+    @Test
+    fun testEncryptDecryptEntropy_shouldMatchOriginal() {
+        // Given
+        val wordCount = Mnemonics.WordCount.COUNT_24
+        val originalEntropy = Mnemonics.MnemonicCode(wordCount).toEntropy()
```
mitsinsar: Same as above

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        // Verify words match
+        assertEquals("Decrypted mnemonic words should match original words",
+            originalWords,
+            decryptedWords
+        )
+    }
+
+    @Test
+    fun testEncryptDecryptEntropy_withEmptyArray() {
+        // Given
```
mitsinsar: there shouldn't be any `given`- `when` - `then` comments in our tests.

We are separating those sections by using new line.

common-sdk/src/main/kotlin/com/algorand/wallet/account/detail/domain/model/AccountType.kt
```diff
@@ -12,23 +12,39 @@
 
 package com.algorand.wallet.account.detail.domain.model
 
-sealed interface AccountType {
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
 
+sealed interface AccountType: Parcelable {
+
+    @Parcelize
```
mitsinsar: We shouldn't use `parcelize` in our common-sdk domain models. 

There are a few models doing that already, and I'll remove them as well. Our component needs to be platform-independent. This also makes things harder when we want to unit test.

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/RecoverWithPassphraseViewModel.kt
```diff
@@ -36,6 +39,10 @@ class RecoverWithPassphraseViewModel @Inject constructor(
 ) : BaseViewModel() {
 
     private val mnemonic: String? = savedStateHandle.getOrElse(MNEMONIC_KEY, null)
+    private val accountType: AccountType = savedStateHandle.getOrElse(
```
mitsinsar: We shouldn't pass `AccountType`. Account type is combination of local account and account information and it has nothing to do with flow

app/src/main/kotlin/com/algorand/android/ui/common/warningconfirmation/BackupInfoViewModel.kt
```diff
@@ -12,21 +12,71 @@
...
 
 @HiltViewModel
 class BackupInfoViewModel @Inject constructor(
-    private val onboardingPassphraseUnderstandEventTracker: OnboardingPassphraseUnderstandEventTracker
+    private val onboardingPassphraseUnderstandEventTracker: OnboardingPassphraseUnderstandEventTracker,
+    private val algoAccountSdk: AlgoAccountSdk,
+    savedStateHandle: SavedStateHandle
 ) : BaseViewModel() {
 
+    private val accountType: AccountType = savedStateHandle.getOrElse(
```
mitsinsar: Same as previous comment

app/src/main/kotlin/com/algorand/android/ui/register/PassphraseValidationViewModel.kt
```diff
@@ -39,4 +46,31 @@ class PassphraseValidationViewModel @Inject constructor(
     fun updateAccountBackupState(publicKey: String, isBackedUp: Boolean) {
         accountManager.updateAccountBackupState(publicKey, isBackedUp)
     }
+
+    fun getMnemonic(args: PassphraseValidationFragmentArgs): String? {
+        val seedId = (args.accountCreation?.type as? AccountCreation.Type.HdKey)?.seedId
+
+        return seedId?.let {
+            runBlocking(Dispatchers.IO) {
+                val entropy = getHdWalletEntropyFromSeedId.invoke(it)
```
mitsinsar: This is a big no. You are blocking UI thread while calling suspend function. These kind of calls needs to be in background threads

app/src/main/kotlin/com/algorand/android/ui/register/PassphraseValidationViewModel.kt
```diff
@@ -39,4 +46,31 @@ class PassphraseValidationViewModel @Inject constructor(
     fun updateAccountBackupState(publicKey: String, isBackedUp: Boolean) {
         accountManager.updateAccountBackupState(publicKey, isBackedUp)
     }
+
+    fun getMnemonic(args: PassphraseValidationFragmentArgs): String? {
+        val seedId = (args.accountCreation?.type as? AccountCreation.Type.HdKey)?.seedId
+
+        return seedId?.let {
```
mitsinsar: Please create a usecase that returns mnemonic. This logic is duplicated and you are using algosdk functions without encapsulating it

app/src/main/kotlin/com/algorand/android/usecase/IsOnHdWalletUseCase.kt
```diff
@@ -0,0 +1,38 @@
...
+                !isProdReleaseVariant()
+        return isHdWalletToggleEnabled
+    }
+
+    fun isConnectedToTestnet(): Boolean {
+        return getIsActiveNodeTestnetUseCase.invoke()
+    }
+
+    fun isProdReleaseVariant(): Boolean {
+        return getIsProductionReleaseUseCase.invoke()
```
mitsinsar, line 36: We already have dedicated usecases, why are we injecting them here instead of injecting them into a class that we need?

app/src/main/res/navigation/login_navigation.xml
mitsinsar: As I mentioned in some of my prvious comments. Using account type as an argument is not a correct implementation.

We are coupling a model that uses local account and account information with account creation arg.

common-sdk/src/androidTest/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
mitsinsar: We can remove this when we remove hilt injection in our test class

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/GetHdWalletEntropyFromSeedIdUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License
+ */
+
+package com.algorand.wallet.account.core.domain.usecase
+
+import com.algorand.wallet.account.local.domain.repository.HdSeedRepository
+import com.algorand.wallet.encryption.SecretKeyEncryptionManager
+import javax.inject.Inject
+
+internal class GetHdWalletEntropyFromSeedIdUseCase @Inject constructor(
```
mitsinsar: This is local account specific usecase. We should move this usecase into local account package

common-sdk/src/main/kotlin/com/algorand/wallet/account/local/domain/repository/HdSeedRepository.kt
```diff
@@ -21,13 +21,23 @@ internal interface HdSeedRepository {
...
+    suspend fun getEncryptedEntropy(seedId: Int): ByteArray?
 
-    suspend fun deleteHdSeed(address: String)
+    suspend fun getHdSeed(encryptedEntropy: ByteArray): HdSeed?
+
+    suspend fun getAllHdSeed(entropyCustomName: String): List<HdSeed>
+
+    fun updateHdSeedCustomNameAsFlow(hdSeed: HdSeed): Flow<Unit>
+
+    fun addHdSeedAsFlow(hdSeed: HdSeed, encryptedEntropy: ByteArray, encryptedSeed: ByteArray): Flow<Unit>
```
mitsinsar: What is the purpose of this function? Why do we need to do this as flow?

common-sdk/src/main/kotlin/com/algorand/wallet/account/local/domain/repository/HdSeedRepository.kt
```diff
@@ -21,13 +21,23 @@ internal interface HdSeedRepository {
...
 
-    suspend fun addHdSeed(seed: HdSeed)
+    suspend fun getEncryptedEntropy(seedId: Int): ByteArray?
 
-    suspend fun deleteHdSeed(address: String)
+    suspend fun getHdSeed(encryptedEntropy: ByteArray): HdSeed?
+
+    suspend fun getAllHdSeed(entropyCustomName: String): List<HdSeed>
+
+    fun updateHdSeedCustomNameAsFlow(hdSeed: HdSeed): Flow<Unit>
```
mitsinsar: What is the purpose of this function? Why do we need to do this as flow?

mitsinsar (changes requested):

common-sdk/src/main/kotlin/com/algorand/wallet/account/detail/domain/model/AccountType.kt
```diff
@@ -12,23 +12,39 @@
...
     companion object {
         fun AccountType.canSignTransaction(): Boolean {
             return this is Algo25 || this is LedgerBle || this is RekeyedAuth
         }
+        fun AccountType.wordCount(): Int {
+            return when (this) {
+                is Algo25 -> 25
+                is HdKey -> 24
+                else -> 0
+            }
```
mitsinsar: This function shouldn't be here. This model is a combination of local account and account info we get from chain. Coupling word count with this model is wrong

common-sdk/src/main/kotlin/com/algorand/wallet/account/detail/domain/model/AccountType.kt
michaeltchuang: we seem to need this though because we're passing `accountType` as a nav graph bundle param between onboarding screens...i'm open to suggestions.

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
michaeltchuang: keep in mind, this is an instrumented test, because I needed AndroidKeyStore.  I wanted to do a unit test at first, but it seem we don't want to add RoboElectric, so let's discuss next week what kind of encryption test we want in this project.  I'll remove this dummy test I was using for my own testing in PR draft though.

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
michaeltchuang: test removed so not relevant anymore

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
michaeltchuang: test file removed so not relevant anymore

common-sdk/src/androidTest/kotlin/com/algorand/wallet/encryption/AESPlatformManagerImplTest.kt
michaeltchuang: test file removed so not relevant anymore

common-sdk/src/main/kotlin/com/algorand/wallet/account/local/domain/repository/HdSeedRepository.kt
michaeltchuang: we need to update the name of wallet record after it's inserted to display in UI.  It doesn't have to be flow but i just need a call back to know it was updated.  For some reason rowsAffected after an update isn't working for me so i'm still investigating

common-sdk/src/main/kotlin/com/algorand/wallet/account/local/domain/repository/HdSeedRepository.kt
michaeltchuang: it's basically the insert function for `HdSeeds` table.  When creating a a new entry, i also need to update the name column afterwards (learning the `seedId` after insert), so I needed a callback.  I was using a flow for now...still investigating if suspend is better

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/GetHdWalletEntropyFromSeedIdUseCase.kt
michaeltchuang: I ended up deleting this use case, it wasn't needed anymore

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/RecoverWithPassphraseViewModel.kt
michaeltchuang: oh do I parcelize `CreateAccount` instead then?  what's difference between `core` and `localaccount` also? 

i was also thinking just storing the generated account into db in beginning and just passing address in the flow (and pulling data from a use case for each step/screen).  More secure so we're not passing private key at all in bundle

app/src/main/kotlin/com/algorand/android/usecase/IsOnHdWalletUseCase.kt
michaeltchuang: you mentioned if we have multiple places that needed the HD feature flag check...to not copy and paste and create a dedicated HD feature flag use case

app/src/main/res/navigation/recover_with_passphrase_navigation.xml
michaeltchuang: I switched to `OnboardingAccountType`

app/src/main/res/navigation/login_navigation.xml
michaeltchuang: i'm using a new model now called `OnboardingAccountType`, app module and parcelize

common-sdk/src/main/kotlin/com/algorand/wallet/account/detail/domain/model/AccountType.kt
michaeltchuang: we're not doing this anymore

common-sdk/src/androidTest/AndroidManifest.xml
michaeltchuang: i removed this

common-sdk/src/main/kotlin/com/algorand/wallet/account/local/domain/repository/HdSeedRepository.kt
michaeltchuang: i changed this back to suspend

app/src/main/kotlin/com/algorand/android/models/OnboardingAccountType.kt
```diff
@@ -0,0 +1,23 @@
...
+    data object HdKey : OnboardingAccountType
+
+    companion object {
+        @Suppress("MagicNumber")
+        fun OnboardingAccountType.wordCount(): Int {
+            return when (this) {
+                is Algo25 -> 25
+                is HdKey -> 24
+            }
+        }
```
mitsinsar: Do we really need this extension function? Extensions are good for adding new ability to classes/interfaces we don't have control over. Class below does the same thing;

```
sealed interface OnboardingAccountType : Parcelable {

    val wordCount: Int
    
    @Parcelize
    data object Algo25 : OnboardingAccountType {
        override val wordCount: Int = 25
    }

    @Parcelize
    data object HdKey : OnboardingAccountType {
        override val wordCount: Int = 24
    }
}
```
No need to change this if you don't need to send a commit for something else, but if you do, I think not using extension for these cases would be a better approach.

app/src/main/kotlin/com/algorand/android/usecase/RegisterIntroPreviewUseCase.kt
```diff
@@ -15,22 +15,26 @@ package com.algorand.android.usecase
...
     private val accountManager: AccountManager,
     private val registerIntroPreviewMapper: RegisterIntroPreviewMapper,
-    private val registerIntroFragmentEventTracker: RegisterIntroFragmentEventTracker
+    private val registerIntroFragmentEventTracker: RegisterIntroFragmentEventTracker,
+    private val getAllHdSeeds: GetAllHdSeeds
 ) {
 
     fun getRegisterIntroPreview(isShowingCloseButton: Boolean) = flow {
         val hasAccount = accountManager.accounts.value.isNotEmpty()
+        val hasHdWallet = getAllHdSeeds.invoke().isNotEmpty()
```
mitsinsar: All we need to check if there is a row in db or not. If we use this usecase, db will query every row, then mapper will map every item. You should create a new use case to ask if there is an item or not. 
I think you can use `EXISTS (SELECT 1...)` which returns immediately once it finds a matching row.

app/src/main/kotlin/com/algorand/android/usecase/AccountAdditionUseCase.kt
```diff
@@ -60,24 +87,49 @@ class AccountAdditionUseCase @Inject constructor(
...
+            var privateKey = aesPlatformManager.decryptByteArray(type.encryptedPrivateKey)
+            var entropy = aesPlatformManager.decryptByteArray(type.encryptedEntropy)
+            val seedId = addHdSeed(entropy)
+            if (seedId.isSuccess) {
+                // seedId was entered successfully
+                addHdKeyAccount(
+                    address,
+                    type.publicKey,
+                    privateKey,
+                    seedId.getDataOrNull() ?: -1,
```
mitsinsar: This doesn't seem okay. If we get an error, which means we couldn't save somehow, we will assign wrong seedId

app/src/main/kotlin/com/algorand/android/usecase/AccountAdditionUseCase.kt
```diff
@@ -50,8 +57,28 @@ class AccountAdditionUseCase @Inject constructor(
...
             when (this) {
+                is Type.HdKey -> {
+                    var entropy = aesPlatformManager.decryptByteArray(this.encryptedEntropy)
+                    val seedId = addHdSeed(entropy)
+                    if (seedId.isSuccess) {
+                        updateNoAuthAccountToHdKey(
+                            address,
+                            publicKey,
+                            encryptedPrivateKey,
+                            seedId.getDataOrNull() ?: -1,
```
mitsinsar: Same as below

app/src/main/kotlin/com/algorand/android/ui/register/BackupPassphraseFragment.kt
```diff
@@ -76,16 +74,12 @@ class BackupPassphraseFragment : DaggerBaseFragment(R.layout.fragment_backup_pas
...
-            } catch (exception: Exception) {
-                navBack()
-            }
-        } ?: run { navBack() }
+        try {
+            val mnemonic = backupPassphraseViewModel.getMnemonic(args) ?: throw Exception("Mnemonic cannot be null.")
+            binding.passphraseBoxView.setPassphrases(mnemonic)
+        } catch (exception: Exception) {
+            navBack()
+        }
```
mitsinsar: Since we have try catch in `getMnemonic` function, we don't need to use try catch here.

Also, using try catch is way costly compared to simple null check. If it is null, jvm needs to capture stack trace, unwind the call stack and then jump to the catch block. Can you change it to null check please

app/src/main/kotlin/com/algorand/android/ui/register/BackupPassphraseViewModel.kt
```diff
@@ -35,4 +41,27 @@ class BackupPassphraseViewModel @Inject constructor(
...
+        } ?: run {
+            val encryptedAlgo25Key = (args.accountCreation?.type as? AccountCreation.Type.Algo25)?.encryptedSecretKey
+
+            val secretKey = encryptedAlgo25Key?.let {
+                aesPlatformManager.decryptByteArray(encryptedAlgo25Key)
+            } ?: getAccountSecretKey(args.publicKeyOfAccountToBackup)
+
+            secretKey?.let {
+                try {
+                    val mnemonic = Sdk.mnemonicFromPrivateKey(it) ?: throw Exception("Mnemonic cannot be null.")
```
mitsinsar, line 59: Can you use our new sdk interface please

app/src/main/kotlin/com/algorand/android/ui/register/BackupPassphraseViewModel.kt
```diff
@@ -35,4 +41,27 @@ class BackupPassphraseViewModel @Inject constructor(
     fun getAccountSecretKey(publicKey: String): ByteArray? {
         return accountManager.getAccount(publicKey)?.getSecretKey()
     }
+
+    fun getMnemonic(args: BackupPassphraseFragmentArgs): String? {
+        val encryptedEntropy = (args.accountCreation?.type as? AccountCreation.Type.HdKey)?.encryptedEntropy
```
mitsinsar, line 46: There is so many nested run, let, try catch functions here.

Can you please split this function into smaller functions and make it readable?

app/src/main/kotlin/com/algorand/android/ui/register/PassphraseValidationViewModel.kt
```diff
@@ -39,4 +45,27 @@ class PassphraseValidationViewModel @Inject constructor(
...
+            secretKey?.let {
+                try {
+                    val mnemonic = Sdk.mnemonicFromPrivateKey(it) ?: throw Exception("Mnemonic cannot be null.")
+                    mnemonic
+                } catch (exception: Exception) {
+                    null
+                }
+            } ?: run { null }
+        }
+    }
```
mitsinsar, line 70: This seems like duplication of BackupPassphraseViewModel function. Can you extract them into a usecase

app/src/main/kotlin/com/algorand/android/ui/compose/widget/PeraCard.kt
```diff
@@ -0,0 +1,119 @@
...
+@Composable
+fun PeraCard(
+    title: String,
+    description: String,
+    footer: String,
+    highlighted: String? = null,
+    onClick: () -> Unit
+) {
+    OutlinedCard(
+        modifier = Modifier
```
mitsinsar, line 48: Can you please move Modifier into args so that we can set its attributes

app/src/main/kotlin/com/algorand/android/ui/compose/widget/GroupChoiceWidget.kt
```diff
@@ -0,0 +1,103 @@
...
+fun ItemChoiceWidget(
+    modifier: Modifier = Modifier,
+    title: String,
+    description: String,
+    icon: ImageVector,
+    iconContentDescription: String,
+    onClick: () -> Unit
+) {
+    Row(
+        modifier = modifier.then(Modifier
```
mitsinsar: Do we need to do that? Can't we use modifier instead? Seems like no need to create new modifier

app/src/main/kotlin/com/algorand/android/ui/common/warningconfirmation/BackupInfoViewModel.kt
```diff
@@ -12,21 +12,76 @@
...
     fun logOnboardingIUnderstandClickEvent() {
         viewModelScope.launch {
             onboardingPassphraseUnderstandEventTracker.logOnboardingPassphraseUnderstandEvent()
         }
     }
+
+    fun createAccount(): AccountCreation {
+        if (onboardingAccountType == OnboardingAccountType.HdKey) {
+            val account = algoAccountSdk.createHdAccount()
+                ?: throw IllegalArgumentException("Failed to create Bip39 account")
```
mitsinsar: Please don't throw exception if there is something null. We should handle it properly. This leads to crash

app/src/main/kotlin/com/algorand/android/ui/common/warningconfirmation/BackupInfoViewModel.kt
```diff
@@ -12,21 +12,76 @@
...
+                    aesPlatformManager.encryptByteArray(account.privateKey),
+                    aesPlatformManager.encryptByteArray(account.entropy),
+                    account.account,
+                    account.change,
+                    account.keyIndex,
+                    account.derivationType,
+                ),
+                creationType = CreationType.CREATE
+            )
+        } else {
```
mitsinsar: Can you split this function into two functions please

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
```diff
@@ -79,6 +84,22 @@ class NameRegistrationViewModel @Inject constructor(
...
+        return accountType is AccountCreation.Type.HdKey
+    }
+
+    fun getWalletId(callback: (Int) -> Unit) {
+        viewModelScope.launch {
+            val maxSeedId: Int? = getMaxHdSeedId.invoke()
+            val walletId = (maxSeedId ?: 0) + 1
+            callback(walletId)
+        }
+    }
```
mitsinsar: Don't do this please. You should use `observable` for these kind of operations

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/usecase/RecoverWithPassphrasePreviewUseCase.kt
```diff
@@ -121,9 +128,14 @@ class RecoverWithPassphrasePreviewUseCase @Inject constructor(
...
     }
 
+    @SuppressWarnings("MagicNumber")
     fun validateEnteredMnemonics(preview: RecoverWithPassphrasePreview) = flow {
         try {
             emit(preview.copy(showLoadingDialogEvent = Event(Unit)))
+            // faking the duration of this process for some time before moving forward
+            // so the loading animation is effective.
+            // After deleting it delete also the @SuppressWarnings("MagicNumber")
+            delay(2500L)
```
mitsinsar: Could you please remove this before merging

mitsinsar (changes requested):

common-sdk/src/main/kotlin/com/algorand/wallet/account/detail/domain/model/AccountType.kt
michaeltchuang: this has been fixed in another PR, we're using `OnboardingAccountType` now

app/src/main/kotlin/com/algorand/android/models/OnboardingAccountType.kt
michaeltchuang: changed to this

app/src/main/kotlin/com/algorand/android/usecase/RegisterIntroPreviewUseCase.kt
michaeltchuang: i created the function `getSeedIdIfExistingEntropy`

app/src/main/kotlin/com/algorand/android/usecase/AccountAdditionUseCase.kt
michaeltchuang: there was already a check handling success/error above this but i modified this code so hopefully it's more understandable now

app/src/main/kotlin/com/algorand/android/usecase/AccountAdditionUseCase.kt
michaeltchuang: same point as above

app/src/main/kotlin/com/algorand/android/ui/common/warningconfirmation/BackupInfoViewModel.kt
michaeltchuang: i split these to two now

app/src/main/kotlin/com/algorand/android/ui/common/warningconfirmation/BackupInfoViewModel.kt
michaeltchuang: there was already a a try / catch on function calling this, but i return null and just navback() now

app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/usecase/RecoverWithPassphrasePreviewUseCase.kt
michaeltchuang: i removed it now we know it's working...gray on light theme might be slightly off but i think we can fix that after launch

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
michaeltchuang: i fixed it by moving this data into preview object

app/src/main/kotlin/com/algorand/android/ui/register/BackupPassphraseFragment.kt
michaeltchuang: good catch, fixed

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
```diff
@@ -42,15 +47,37 @@ class NameRegistrationViewModel @Inject constructor(
     private val accountCreation = savedStateHandle.get<AccountCreation>(ACCOUNT_CREATION_KEY)
     private val accountAddress = accountCreation?.address
     private val accountName = accountCreation?.customName
+    protected val accountType = accountCreation?.type
```
mitsinsar, line 50: This can be private

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
```diff
@@ -42,15 +47,37 @@ class NameRegistrationViewModel @Inject constructor(
...
     val predefinedAccountName: String
         get() = accountName.takeUnless { it.isNullOrBlank() } ?: accountAddress.toShortenedAddress()
 
+    init {
+        viewModelScope.launch {
+            updatePreviewWithHdWalletData()
+        }
+    }
+
+    fun updatePreviewWithHdWalletData() {
```
mitsinsar, line 62: This can be private

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
```diff
@@ -42,15 +47,37 @@ class NameRegistrationViewModel @Inject constructor(
...
     private val accountName = accountCreation?.customName
+    protected val accountType = accountCreation?.type
+    private var walletId: Int? = null
 
     val predefinedAccountName: String
         get() = accountName.takeUnless { it.isNullOrBlank() } ?: accountAddress.toShortenedAddress()
 
+    init {
+        viewModelScope.launch {
+            updatePreviewWithHdWalletData()
```
mitsinsar, line 58: Function you call is not suspended. There is no need to create coroutineScope. Function creates its own scope

mitsinsar (approved):

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
michaeltchuang: i'll fix this in my recovery hd wallet PR

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
michaeltchuang: i'll fix this in my recovery hd wallet PR

app/src/main/kotlin/com/algorand/android/ui/register/nameregistration/NameRegistrationViewModel.kt
michaeltchuang: i'll fix this in my recovery hd wallet PR

app/src/main/kotlin/com/algorand/android/ui/compose/widget/GroupChoiceWidget.kt
michaeltchuang: fixed

PR #160, opened 2025-02-25
PERA-1694 :: Add in Play Install Referrer tracking

michaeltchuang:
# Pull Request Template

## Description
- Add in Play Install Referrer tracking
- Now app will pass Google Analytics 4 UTM variables in FirebaseAnalytics.logEvent()

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1694

## Checklist

## Additional Notes
- I created a GA4 utm guide here - https://solas.algorand.foundation/doc/google-analytics-referral-tracking-guide-vUJHwihGFR
- If you want to test this...put a breakpoint in install referrer helper class and install app from Android Studio (to see referrer url) or maybe try adb approach like this
```
adb shell am broadcast -a com.android.vending.INSTALL_REFERRER \
-n com.algorand.android.debug/com.algorand.android.debug.referrer.InstallReferrerReceiver \
--es "referrer" "utm_source=pera_website&utm_medium=referral&utm_campaign=testing"
```


mitsinsar (changes requested):
Existing implementation is not testable. Since this is not hot-fix that we need to publish today, I think we have enough time to add its test.

mitsinsar:
Existing class takes context as parameter. That makes it impossible to write unit test.

Also a function needs to do one thing. Right now, `InstallReferrerHelper:fetchInstallReferrer` function;
- Creates client
- Observers its responses
- Decodes url queries
- Writes referrer data into shared preferences

michaeltchuang:
> Existing implementation is not testable. Since this is not hot-fix that we need to publish today, I think we have enough time to add its test.

I added a unit test

> Existing class takes context as parameter. That makes it impossible to write unit test.
> 
> Also a function needs to do one thing. Right now, `InstallReferrerHelper:fetchInstallReferrer` function;
> 
> * Creates client
> * Observers its responses
> * Decodes url queries
> * Writes referrer data into shared preferences

I added a unit test using roboelectric.  I figure that should work for us in terms of test coverage?

I'm also trying not to over-engineer things either so we can get this fix out quickly.   Are you thinking each of these steps should be use cases instead?

app/src/main/kotlin/com/algorand/android/MainViewModel.kt
```diff
@@ -299,4 +304,10 @@ class MainViewModel @Inject constructor(
     fun hasAccountAuthority(accountAddress: String): Boolean {
         return accountStateHelperUseCase.hasAccountAuthority(accountAddress)
     }
+
+    fun fetchInstallReferrer() {
+        CoroutineScope(coroutineDispatcher).launch {
```
mitsinsar: Instead of creating coroutine scope, we should use `viewModelScope.launchIO` in ViewModels.

app/src/main/kotlin/com/algorand/android/modules/tracking/core/FirebaseEventTracker.kt
```diff
@@ -14,19 +14,30 @@ package com.algorand.android.modules.tracking.core
...
+import com.algorand.wallet.analytics.domain.utils.UrlReferrerParser.Companion.UTM_SOURCE
+import com.algorand.wallet.analytics.domain.utils.UrlReferrerParser.Companion.UTM_TERM
 import com.google.firebase.analytics.FirebaseAnalytics
+import javax.inject.Inject
 
-class FirebaseEventTracker(
-    private val firebaseAnalytics: FirebaseAnalytics
+class FirebaseEventTracker @Inject constructor (
+    private val firebaseAnalytics: FirebaseAnalytics,
+    private val referrerRepository: ReferrerRepository
```
mitsinsar: By adding referrer repository into tracker, you are coupling tracker and referrer components. 

You should create payload and pass it to `logEvent` function. That means, there shouldn't be any change in this class.

Also, to get saved referred data, instead of injecting repository, you should use dedicated use case. Otherwise you reach data layer directly.

app/src/main/kotlin/com/algorand/android/utils/walletconnect/WalletConnectFirebaseEventLogger.kt
```diff
@@ -12,21 +12,27 @@
...
 
 /**
  * This class will be provided by hilt - AppModule.kt
  */
 class WalletConnectFirebaseEventLogger(
-    private val firebaseAnalytics: FirebaseAnalytics,
-    private val algodInterceptor: AlgodInterceptor
+    private val algodInterceptor: AlgodInterceptor,
+    private val firebaseEventTracker: FirebaseEventTracker,
+    private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.IO
```
mitsinsar: We shouldn't create coroutine scope like this. This coroutine scope is not attached to any lifecycle. That means, it can stay alive even if VM or Activity is destroyed. If you want to log something, function that calls `logEvent` needs to be suspended.

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/InstallReferrerApiClient.kt
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License
+ */
+
+package com.algorand.wallet.analytics.data
+
+import android.content.Context
+import com.android.installreferrer.api.InstallReferrerClient
+import javax.inject.Inject
+
+class InstallReferrerApiClient @Inject constructor(
```
mitsinsar: If you create client class like this, you still need context in common-sdk. 

You should create your builder interface in common-sdk and then create implementation of this interface in `app` module. 

Also, you are providing this class as Singleton which has context in its constructor. Referrer api does it jobs, and it's done. We shouldn't keep its instance in memory.

One more thing, builders doesn't require unit test.

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/ReferrerManagerImpl.kt
```diff
@@ -0,0 +1,64 @@
...
+internal class ReferrerManagerImpl @Inject constructor(
+    private val installReferralApiClient: InstallReferrerApiClient,
+    private val referrerRespository: ReferrerRepository,
+    private val urlQueryParser: UrlReferrerParser,
+    private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.IO
+) : ReferrerManager {
+
+    override suspend fun initialize() {
+        val installReferrerClient = installReferralApiClient.initialize()
+        installReferrerClient.startConnection(object : InstallReferrerStateListener {
```
mitsinsar: You can create an interface for client with suspend function that returns String, and use the result here. 

```
interface PeraReferralClient {
    suspend fun getReferrerUrl(): String?
}

internal class ReferralManagerImpl(
    private val peraReferralClient: PeraReferralClient,
    ...
) : ReferralManager {

    override suspend fun initialize() {
        val referrerUrl = peraReferralClient.getReferrerUrl()
        ...
    }
}
```

Because if you inject third party library client directly, you are not abstracting anything

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/domain/repository/ReferrerRepository.kt
```diff
@@ -0,0 +1,20 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+
+package com.algorand.wallet.analytics.domain.repository
+
+import com.algorand.wallet.analytics.domain.usecases.model.ReferralData
+
+interface ReferrerRepository{
```
mitsinsar: This needs to be internal.

UseCases should be the only public interfaces

common-sdk/src/test/kotlin/com/algorand/wallet/analytics/InstallReferrerApiClientTest.kt
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+
+package com.algorand.wallet.analytics.data
+
+import android.content.Context
```
mitsinsar: Unit tests shouldn't contain android framework classes

common-sdk/src/test/kotlin/com/algorand/wallet/analytics/ReferrerManagerImpTest.kt
```diff
@@ -0,0 +1,144 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+
+package com.algorand.wallet.analytics.data
+
+import android.util.Log
```
mitsinsar: Same as above

common-sdk/src/test/kotlin/com/algorand/wallet/analytics/UrlReferrerParserTest.kt
```diff
@@ -0,0 +1,148 @@
...
+import java.nio.charset.StandardCharsets
+
+class UrlReferrerParserTest {
+
+    private lateinit var sut: UrlReferrerParser
+
+    @Before
+    fun setUp() {
+        sut = UrlReferrerParser()
+        mockkStatic(URLDecoder::class)
```
mitsinsar: Places we use mockkStatic, usually violates some stuff. When you use mockkStatic, you modify its behavior globally, which may lead race conditions and flaky tests. When you use interface, you isolate your test from everything else.

app/src/main/kotlin/com/algorand/android/modules/tracking/core/FirebaseEventTracker.kt
```diff
@@ -60,6 +71,18 @@ class FirebaseEventTracker(
...
+    suspend fun addReferralDataToBundle(bundle: Bundle): Bundle {
+        val referralData = referrerRepository.getReferrerData()
+        referralData.let {
+            it.utmSource?.let { source -> bundle.putString(UTM_SOURCE, source) }
+            it.utmMedium?.let { medium -> bundle.putString(UTM_MEDIUM, medium) }
+            it.utmCampaign?.let { campaign -> bundle.putString(UTM_CAMPAIGN, campaign) }
+            it.utmTerm?.let { term -> bundle.putString(UTM_TERM, term) }
+            it.utmContent?.let { content -> bundle.putString(UTM_CONTENT, content) }
+        }
+        return bundle
```
mitsinsar: As I mentioned, this is not event tracker's job. Event tracker, tracks event you give. It doesn't add anything into those bundles. 

Whenever you call `logEvent` it is going to add this payload, even if you just want to track a transaction, or screen...

mitsinsar (changes requested):
I haven't reviewed tests yet. We need to make sure our prod code is valid, and then I will. check tests

app/src/main/kotlin/com/algorand/android/modules/tracking/core/FirebaseEventTracker.kt
michaeltchuang: > As I mentioned, this is not event tracker's job. Event tracker, tracks event you give. It doesn't add anything into those bundles.
> 
> Whenever you call `logEvent` it is going to add this payload, even if you just want to track a transaction, or screen...

That's what we want to happen though, every analytics event (screen view, event, etc) going up should have corresponding referrer attribution info (e.g. in our case it's GA4 UTM variables).

mitsinsar: Then we need to create another event tracker with specific name. Because this class name says nothing about referral. We can have `PeraEventTrackerImpl`,  (since eveeeery tracking needs these referral data) that implements `PeraEventTracker`. In the new tracker, you can get referral data, and pass it to firebase event tracker. 

When you replace dependency injections, classes that uses this interface won't be affected.

app/src/main/kotlin/com/algorand/android/MainViewModel.kt
michaeltchuang: fixed

common-sdk/src/test/kotlin/com/algorand/wallet/analytics/UrlReferrerParserTest.kt
michaeltchuang: removed

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/ReferrerManagerImpl.kt
michaeltchuang: i abstracted it into app module

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/domain/repository/ReferrerRepository.kt
michaeltchuang: fixed

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/InstallReferrerApiClient.kt
michaeltchuang: i think I addressed this by moving build to app module...take a look at what i did if that's what you were thinking

app/src/main/kotlin/com/algorand/android/utils/walletconnect/WalletConnectFirebaseEventLogger.kt
michaeltchuang: I switched to suspend

app/src/main/kotlin/com/algorand/android/modules/tracking/core/FirebaseEventTracker.kt
michaeltchuang: I switched to use case

app/src/main/kotlin/com/algorand/android/dependencyinjection/AppModule.kt
```diff
@@ -187,4 +190,14 @@ object AppModule {
     fun provideNotificationManager(@ApplicationContext appContext: Context): NotificationManager? {
         return appContext.getSystemService(Context.NOTIFICATION_SERVICE) as? NotificationManager
     }
+
+    @Provides
+    @Singleton
+    fun provideContext(application: Application): Context {
```
mitsinsar: There is already context provider, so no need to create this one. You can check the function above `provideNotificationManager`. This is how you can inject context to functions

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/domain/util/Constants.kt
```diff
@@ -0,0 +1,9 @@
+package com.algorand.wallet.analytics.domain.util
+
+object GA4 {
```
mitsinsar: Could you please make this internal

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/domain/usecase/GetReferrerDataUseCase.kt
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License
+ */
+
+package com.algorand.wallet.analytics.domain.usecase
+
+import com.algorand.wallet.analytics.domain.repository.ReferrerRepository
+import com.algorand.wallet.analytics.domain.usecases.model.ReferrerData
+import javax.inject.Inject
+
+internal class GetReferrerDataUseCase @Inject constructor(
```
mitsinsar: You don't need to create this use case, because it has no logic. What you can do is,
instead of providing use cases like below;
```
     @Provides
     fun provideGetReferrerData(useCase: GetReferrerDataUseCase): GetReferrerData = useCase

     @Provides
     fun provideSaveReferrerData(useCase: SaveReferrerDataUseCase): SaveReferrerData = useCase
```

You can remove usecase implementation classes and provide interface as below;

```
     @Provides
     fun provideGetReferrerData(repository: ReferrerRepository): GetReferrerData = GetReferrerData(repository::getReferrerData)

     @Provides
     fun provideSaveReferrerData(repository: ReferrerRepository): SaveReferrerData = SaveReferrerData(repository::saveReferrerData)
```

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/domain/usecase/SaveReferrerDataUseCase.kt
```diff
@@ -0,0 +1,14 @@
+package com.algorand.wallet.analytics.domain.usecase
+
+import com.algorand.wallet.analytics.domain.repository.ReferrerRepository
+import com.algorand.wallet.analytics.domain.usecases.model.ReferrerData
+import javax.inject.Inject
+
+internal class SaveReferrerDataUseCase @Inject constructor(
```
mitsinsar: as commented above

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/service/PeraEventTrackerImpl.kt
```diff
@@ -60,12 +72,24 @@ class FirebaseEventTracker(
         }
     }
 
+    suspend fun addReferralDataToBundle(bundle: Bundle): Bundle {
```
mitsinsar: This function needs to be private

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/service/ReferrerQueryParamParserImpl.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.algorand.wallet.analytics.domain.util.GA4.UTM_CAMPAIGN
+import com.algorand.wallet.analytics.domain.util.GA4.UTM_CONTENT
+import com.algorand.wallet.analytics.domain.util.GA4.UTM_MEDIUM
+import com.algorand.wallet.analytics.domain.util.GA4.UTM_SOURCE
+import com.algorand.wallet.analytics.domain.util.GA4.UTM_TERM
+import java.net.URLDecoder
+import java.nio.charset.StandardCharsets
+import javax.inject.Inject
+
+class ReferrerQueryParamParserImpl @Inject constructor(): ReferrerQueryParamParser {
```
mitsinsar: Can you make this internal please

common-sdk/build.gradle.kts
```diff
@@ -28,15 +28,21 @@ android {
...
+    }
 }
 
 dependencies {
 
     api(libs.algosdk)
     api(libs.algorand.go.mobile)
+    api(libs.xhdwalletapi)
+    api(libs.kotlin.bip39)
+    api(libs.installreferrer)
```
mitsinsar: Since we provide this library from app module, we don't need this dependency in `common-sdk`. you can move it to app gradle

app/src/main/kotlin/com/algorand/android/MainViewModel.kt
```diff
@@ -299,4 +303,13 @@ class MainViewModel @Inject constructor(
     fun hasAccountAuthority(accountAddress: String): Boolean {
         return accountStateHelperUseCase.hasAccountAuthority(accountAddress)
     }
+
+    fun fetchInstallReferrer() {
+        viewModelScope.launch(Dispatchers.IO) {
+            val referrerUrl = referrerClient.getReferrerUrl()
+            referrerUrl?.let { url ->
+                referrerManager.saveReferrerData(url)
+            }
```
mitsinsar: If we do this here, there no point of having referrerManager. It basically becomes `SaveReferrerData` usecase. 

May I ask you to move this logic into referrerManager class;

```
interface ReferrerManager {
    suspend fun fetchInstallReferrer()
}
```

after doing that, you can move lines 309-311 into referrer manager, helps us to delegate this logic into manager and keep one less logic in this MainViewModel

app/src/main/kotlin/com/algorand/android/utils/walletconnect/WalletConnectFirebaseEventLogger.kt
```diff
@@ -12,27 +12,28 @@
...
+import com.algorand.wallet.analytics.domain.service.PeraEventTracker
 
 /**
  * This class will be provided by hilt - AppModule.kt
  */
 class WalletConnectFirebaseEventLogger(
-    private val firebaseAnalytics: FirebaseAnalytics,
-    private val algodInterceptor: AlgodInterceptor
+    private val algodInterceptor: AlgodInterceptor,
+    private val firebaseEventTracker: PeraEventTracker
```
mitsinsar: I think we can rename this as `peraEventTracker`

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/service/ReferrerManagerImpl.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package com.algorand.wallet.analytics.data.service
+
+import com.algorand.wallet.analytics.domain.repository.ReferrerRepository
+import com.algorand.wallet.analytics.domain.service.ReferrerManager
+import com.algorand.wallet.analytics.domain.service.ReferrerQueryParamParser
+import javax.inject.Inject
+
+internal class ReferrerManagerImpl @Inject constructor(
+    private val referrerRespository: ReferrerRepository,
```
mitsinsar: Here, you should use `saveReferrerData: SaveReferrerData` use case, instead of injecting repository.

mitsinsar (changes requested):
Could you please update your test names as well, all the test we have follow the pattern below;
`EXPECT ... WHEN ...`

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/service/ReferrerManagerImpl.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/utils/walletconnect/WalletConnectFirebaseEventLogger.kt
michaeltchuang: good catch, fixed

app/src/main/kotlin/com/algorand/android/MainViewModel.kt
michaeltchuang: fixed

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/service/ReferrerQueryParamParserImpl.kt
michaeltchuang: fixed

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/data/service/PeraEventTrackerImpl.kt
michaeltchuang: fixed

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/domain/usecase/GetReferrerDataUseCase.kt
michaeltchuang: fixed

common-sdk/src/main/kotlin/com/algorand/wallet/analytics/domain/util/Constants.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/dependencyinjection/AppModule.kt
michaeltchuang: fixed

mitsinsar (approved):

PR #191, opened 2025-03-11
PERA-1762 :: Fix view passphrase for account-refactor epic branch

michaeltchuang:
# Pull Request Template

## Description
- Fix view passphrase for account-refactor epic branch

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1762

## Checklist
- [X] Have you tested your changes locally?
- [X] Have you reviewed the code for any potential issues?
- [X] Have you documented any necessary changes in the project's documentation?
- [X] Have you added any necessary tests for your changes?
- [X] Have you updated any relevant dependencies?

## Additional Notes
- Add any additional notes or comments that may be helpful for reviewers.

PR #195, opened 2025-03-12
PERA-1762 :: Migrate view passphrase fragment

mitsinsar:
# Pull Request Template

## Description
- Moved ViewModel into common-sdk and added its test
- Moved getting mnemonic logic into common-sdk and added its test

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1762

## Checklist

## Success state (not really helpful since we block screen recording)

https://github.com/user-attachments/assets/8049b9d5-1327-4422-8aab-a3bf84e30173

## Error state

https://github.com/user-attachments/assets/9819e6ee-9794-42cb-8caf-5a82accf46f3


app/src/main/kotlin/com/algorand/android/modules/viewpassphrase/view/ViewPassphraseFragment.kt
```diff
@@ -78,4 +93,24 @@ class ViewPassphraseFragment : DaggerBaseFragment(R.layout.fragment_view_passphr
...
+    private fun showMnemonics(mnemonicWords: List<String>) {
+        binding.progressBar.hide()
+        binding.passphraseBoxView.apply {
+            setPassphrases(mnemonicWords)
+            show()
+        }
+    }
+
+    private companion object {
+        const val PUBLIC_KEY = "publicKey"
```
michaeltchuang: this should be changed to `address`

app/src/main/kotlin/com/algorand/android/modules/viewpassphrase/view/ViewPassphraseFragment.kt
```diff
@@ -48,27 +56,34 @@ class ViewPassphraseFragment : DaggerBaseFragment(R.layout.fragment_view_passphr
...
+    private fun initObserver() {
+        collectLatestOnLifecycle(viewPassphraseViewModel.viewEvent, viewEventCollector, Lifecycle.State.CREATED)
+        collectLatestOnLifecycle(viewPassphraseViewModel.state, viewStateCollector)
     }
 
     override fun onResume() {
         super.onResume()
         view?.viewTreeObserver?.addOnWindowFocusChangeListener(onWindowFocusChangeListener)
         activity?.disableScreenCapture()
+        val address = requireArguments().getString(PUBLIC_KEY).orEmpty()
```
michaeltchuang: i think this is a typo, if we're using address, let's use address and not public key which will get confusing with hd keys (since both are columns in table)

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+
+@HiltViewModel
+class ViewPassphraseViewModel @Inject constructor(
+    private val getAccountMnemonic: GetAccountMnemonic,
+    private val stateDelegate: StateDelegate<ViewState>,
+    private val eventDelegate: EventDelegate<ViewEvent>
+) : ViewModel(), StateViewModel<ViewState> by stateDelegate, EventViewModel<ViewEvent> by eventDelegate {
+
+    init {
+        stateDelegate.setDefaultState(Idle)
```
michaeltchuang: default state can just be `loading` and we can remove `idle`

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    private fun displayError() {
+        viewModelScope.launch {
+            eventDelegate.sendEvent(ViewEvent.ShowGenericError)
+            eventDelegate.sendEvent(ViewEvent.NavigateBack)
+        }
+    }
+
+    sealed interface ViewState {
+        data object Idle : ViewState
```
michaeltchuang: default state can just be `loading` so we can remove `idle`, but i think we should also add in `error` state.

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private fun displayError() {
+        viewModelScope.launch {
+            eventDelegate.sendEvent(ViewEvent.ShowGenericError)
+            eventDelegate.sendEvent(ViewEvent.NavigateBack)
+        }
+    }
+
+    sealed interface ViewState {
```
michaeltchuang: i think we also should have an abstract BaseViewState (add to ViewModel.kt in common-sdk) that has loading, content, and error so every screen should implement those states if you get what i mean

maybe something like

```
abstract class BaseViewState<T> {
    abstract val isLoading: Boolean
    abstract val content: T?
    abstract val error: String?
    
    val hasError: Boolean get() = error != null
    val hasContent: Boolean get() = content != null
}
```

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
michaeltchuang, line 1: i think i'd prefer if we group this in a section area...so let's create a folder accountdetails that you put viewpassphrase in.  Other "common" sections i want to create would be onboarding (create/recover), transactionsigning, and maybe swap (UI stuff that other folks can use if they want) this year

also we can move the compose widgets here if we have ui folder (in a separate PR)

app/src/main/kotlin/com/algorand/android/modules/viewpassphrase/view/ViewPassphraseFragment.kt
mitsinsar: On my branch, this is still "publicKey"

app/src/main/kotlin/com/algorand/android/modules/viewpassphrase/view/ViewPassphraseFragment.kt
mitsinsar: I agree, I started to use `address` instead of `publicKey` but didn't updated key to create extra work, since purpose of this to implement states

michaeltchuang (changes requested):
i think you're missing key `accountAddress` changes from my PR in

[AccountOptionsBottomSheet.kt](https://github.com/perawallet/pera-android/pull/191/files#diff-49db29851b8df3a85efc71419e4fc813d97024cb6af42282c60a9b81abbdc590)

[AccountOptionsPreviewUseCase.kt](https://github.com/perawallet/pera-android/pull/191/files#diff-c02cf9874e8e4aab8beba5ebba7722ddf90de425c834077d4d56dc8fe3d59eb9)

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
mitsinsar: By setting the default state as Loading, we are making an assumption that an action has already started, which is not the case. State should reflect the true state of the UI.

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
mitsinsar: Reason of creating Idle is to have `not-null` state on the UI. 

Also, in our flow, we navigate back to previous screen when we get an error. So this becomes an `event` instead of a `state`

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
mitsinsar: We shouldn't use `Base` class at all. There is no value of doing that. 

Additionally, loading, error and content states needs to be different classes, instead of booleans. When you keep them as boolean, state management becomes harder. Also, this approach cause us to have bugs because you can have `loading` and `error` at the same time.

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+        stateDelegate.setDefaultState(Idle)
+    }
+
+    fun initViewState(accountAddress: String) {
+        stateDelegate.onState<Idle> {
+            stateDelegate.updateState { ViewState.Loading }
+            viewModelScope.launch {
+                getAccountMnemonic(accountAddress).use(
+                    onSuccess = ::displayMnemonic,
+                    onFailed = { _, _ -> displayError() }
```
michaeltchuang: i think would change to `ViewState.error`

mitsinsar: In our case, we navigate back, we don't display error state. That's why it becomes `ViewEvent` not `ViewState`

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    sealed interface ViewState {
+        data object Idle : ViewState
+        data object Loading : ViewState
+        data class Content(val mnemonicWords: List<String>) : ViewState
+    }
+
+    sealed interface ViewEvent {
```
michaeltchuang: i'm also wondering if we should do same thing here for events...an abstract where every screen should have a navBack

mitsinsar: If we need this kind of logic, we should use composition, instead of inheritance.  Please check `composition over inheritance`

app/src/main/kotlin/com/algorand/android/modules/viewpassphrase/view/ViewPassphraseFragment.kt
michaeltchuang: that's what i referring to in my feedback...we should change this when calling into navgraph to `accountAddress` or something similar

app/src/main/kotlin/com/algorand/android/modules/viewpassphrase/view/ViewPassphraseFragment.kt
michaeltchuang: oh ok...i have the references in my PR...it's like 2 or 3 spots, so quick fixes other than states...no biggie

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
michaeltchuang: i think starting in loading state is fine so user knows we're doing something, since we're going to be fetching content right away.  Otherwise we'll have an extra state that is there a few milliseconds, and it's one less state we need to worry about.  I feel it's over-engineering

common-sdk/src/main/kotlin/com/algorand/wallet/ui/viewpassphrase/viewmodel/ViewPassphraseViewModel.kt
michaeltchuang: The reason we commonly want to use error state (and then the state handling can call `displayError` or equivalent) is because when the previous state is loading...if navback function fails, user is stuck in loading view forever.  I think we can also show something on screen (and turn off loader) while we nav back

Also, we can also count in analytics how many users landed into an error state on ViewPassphrase screen.

mitsinsar: I understand what you said, but our current implementation is about navigating back. If we decide to display error state, then we can update accordingly. But right now, we are trying to solve a problem that doesn't exist. 

Also, we can still track this with this approach. you can simply add your logging into displayError function in current implementation.

michaeltchuang (approved):
I caught a bug in the use case for how you were getting mnemonic for hd...and fixed the test as well.  

I also ran into existing `accountAddress` variable in `AccountOptionsViewModel` so i ended up changing many more places from `publicKey` to `accountAddress`.  I only fixed this in immediate references, but i think we should search all and replace `publicKey` with `accountAddress` for more locations when we have time

PR #201, opened 2025-03-13
PERA-1581 :: Add fetching HD registered addresses code into common sdk

michaeltchuang:
# Pull Request Template

## Description
- Add fetching HD registered addresses code into common sdk

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1581

## Checklist

## Additional Notes
- This is just the common-sdk code so UI code PR for account scanning will be much smaller
- Kover class test coverage is up to 60% now after this PR, maybe we should bump it up too?


common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupFetchHelperImpl.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import com.algorand.wallet.account.info.data.model.AccountFastLookupResponse
+import com.algorand.wallet.account.info.data.service.AccountFastLookupApiService
+import com.algorand.wallet.foundation.PeraResult
+import com.algorand.wallet.foundation.network.utils.request
+import javax.inject.Inject
+
+internal class AccountFastLookupFetchHelperImpl @Inject constructor(
+    private val api: AccountFastLookupApiService
+) : AccountFastLookupFetchHelper {
```
mitsinsar: Can we rename this as `AccountFastLookupRepository`?

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/usecase/AccountInformationUseCases.kt
```diff
@@ -113,6 +114,10 @@ fun interface GetAccountRekeyAdminAddress {
     suspend operator fun invoke(address: String): String?
 }
 
+fun interface GetAccountFastLookup {
+    suspend operator fun invoke(address: String): AccountFastLookup?
```
mitsinsar: This needs to return PeraResult, instead of nullable

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupFetchHelper.kt
```diff
@@ -0,0 +1,20 @@
...
+ * limitations under the License
+ */
+
+package com.algorand.wallet.account.info.data.repository
+
+import com.algorand.wallet.account.info.data.model.AccountFastLookupResponse
+import com.algorand.wallet.foundation.PeraResult
+
+internal interface AccountFastLookupFetchHelper {
+    suspend fun fetchAccountFastLookup(accountAddress: String): PeraResult<AccountFastLookupResponse>
```
mitsinsar: Let's rename this as Repository

Also, repository shouldn't return `Response` model. When you do that, since response model is in data package, your domain layer depends on data layer. Our architecture is domain driven, so this approach doesn't fit. 

You should map response to your domain model in repository

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/model/AccountFastLookupResponse.kt
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License
+ */
+
+package com.algorand.wallet.account.info.data.model
+
+import com.google.gson.annotations.SerializedName
+
+internal data class AccountFastLookupResponse(
+    @SerializedName("algo_value")
+    val algoValue: String,
```
mitsinsar, line 19: I don't know the response details but I assume we can use `BigInteger` for this, instead of String.

and Double for usd value maybe?

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/model/AccountFastLookup.kt
```diff
@@ -0,0 +1,8 @@
+package com.algorand.wallet.account.info.domain.model
+
+data class AccountFastLookup(
+    val algoValue: String,
+    val usdValue: String,
```
mitsinsar: Let's update their type as well

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/usecase/GetAccountFastLookupUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import com.algorand.wallet.account.info.data.mapper.model.AccountFastLookupMapper
+import com.algorand.wallet.account.info.data.repository.AccountFastLookupFetchHelper
+import com.algorand.wallet.account.info.domain.model.AccountFastLookup
+import javax.inject.Inject
+
+internal class GetAccountFastLookupUseCase @Inject constructor(
+    private val accountFastLookupFetchHelper: AccountFastLookupFetchHelper,
+    private val accountFastLookupMapper: AccountFastLookupMapper
+) : GetAccountFastLookup {
```
mitsinsar: When you change return type of this use case, and move mapper into repository, you can remove this usecase implementation class and provide it via hilt like we do; `GetAccountFastLookup(repository::.....)`

common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/PeraBip39SdkImpl.kt
```diff
@@ -12,18 +12,25 @@
...
 import foundation.algorand.xhdwalletapi.Bip32DerivationType
 import foundation.algorand.xhdwalletapi.KeyContext
 import foundation.algorand.xhdwalletapi.XHDWalletAPIAndroid
 import foundation.algorand.xhdwalletapi.XHDWalletAPIBase.Companion.fromSeed
 import foundation.algorand.xhdwalletapi.XHDWalletAPIBase.Companion.getBIP44PathFromContext
 import javax.inject.Inject
 
-internal class PeraBip39SdkImpl @Inject constructor() : PeraBip39Sdk {
+internal class PeraBip39SdkImpl @Inject constructor(
+    private val getAccountFastLookup: GetAccountFastLookup,
```
mitsinsar: Purpose of this interface is to abstract sdk related things, you shouldn't do these operations behind this interface.  Please create a usecase for this operation

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/data/mapper/model/AccountFastLookupMapperImplTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun `EXPECT account does not exist WHEN response has accountExists false`() {
+        val response = AccountFastLookupResponse(
+            algoValue = "0",
+            usdValue = "0.00",
+            calculationType = "estimated",
+            accountExists = false
+        )
+
+        val result = sut.invoke(response)
+
```
mitsinsar, line 69: You can simple create an expected class and then compare these who, instead of comparing their fields one by one.

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/domain/usecase/GetAccountFastLookupUseCaseTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.algorand.wallet.foundation.PeraResult
+import io.mockk.coEvery
+import io.mockk.mockk
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Before
+import org.junit.Test
+import kotlinx.coroutines.test.runTest
+
+class GetAccountFastLookupUseCaseTest {
```
mitsinsar, line 29: Let remove this test when we remove this use case

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/domain/usecase/GetAccountDetailCacheStatusFlowUseCaseTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.test.runTest
+
+class GetAccountDetailCacheStatusFlowUseCaseTest {
```
mitsinsar: Please use test observer instead of `flow.first()`. We need to test every value in it, not just the first one. These tests doesn't test if flow value is updated or not when the value changes.

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/data/repository/AccountInformationCacheHelperImplTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        coEvery { mockAssetHoldingCacheHelper.cacheAssetHolding(address, mockAssetHoldingList) } returns mockAssetHoldings
+        coEvery { mockAccountInformationMapper(mockEntity, mockAssetHoldings) } returns expectedAccountInformation
+
+        val result = sut.cacheAccountInformation(address, mockResponse)
+
+        assertEquals(expectedAccountInformation, result)
+        coVerify(exactly = 1) { mockAccountInformationDao.insert(mockEntity) }
+        coVerify(exactly = 1) { mockAccountInformationErrorCache.remove(address) }
+        coVerify(exactly = 1) { mockAssetHoldingCacheHelper.cacheAssetHolding(address, mockAssetHoldingList) }
+        coVerify(exactly = 1) { mockAccountInformationMapper(mockEntity, mockAssetHoldings) }
```
mitsinsar: No need to verify these two, since you've already mocked these functions with these args.

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/data/repository/AccountInformationCacheHelperImplTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        coEvery { mockAccountInformationEntityMapper(mockResponse) } returns null
+        coEvery { mockAccountInformationDao.isAddressExists(address) } returns false
+
+        val result = sut.cacheAccountInformation(address, mockResponse)
+
+        assertNull(result)
+        coVerify(exactly = 0) { mockAccountInformationDao.insert(any()) }
+        coVerify(exactly = 0) { mockAccountInformationErrorCache.remove(any()) }
+        coVerify(exactly = 1) { mockAccountInformationErrorCache.put(address) }
+        coVerify(exactly = 1) { mockAccountInformationDao.isAddressExists(address) }
```
mitsinsar: No need to verify this, since you've already mocked this function with this arg

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/data/repository/AccountInformationCacheHelperImplTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        coEvery { mockAccountInformationEntityMapper(mockResponse) } returns null
+        coEvery { mockAccountInformationDao.isAddressExists(address) } returns true
+
+        val result = sut.cacheAccountInformation(address, mockResponse)
+
+        assertNull(result)
+        coVerify(exactly = 0) { mockAccountInformationDao.insert(any()) }
+        coVerify(exactly = 0) { mockAccountInformationErrorCache.remove(any()) }
+        coVerify(exactly = 0) { mockAccountInformationErrorCache.put(any()) }
+        coVerify(exactly = 1) { mockAccountInformationDao.isAddressExists(address) }
```
mitsinsar: same as above

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/data/repository/AccountInformationCacheHelperImplTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        coEvery { mockAssetHoldingCacheHelper.cacheAssetHolding(address, emptyList()) } returns emptyAssetHoldings
+        coEvery { mockAccountInformationMapper(mockEntity, emptyAssetHoldings) } returns expectedAccountInformation
+
+        val result = sut.cacheAccountInformation(address, mockResponse)
+
+        assertEquals(expectedAccountInformation, result)
+        coVerify(exactly = 1) { mockAccountInformationDao.insert(mockEntity) }
+        coVerify(exactly = 1) { mockAccountInformationErrorCache.remove(address) }
+        coVerify(exactly = 1) { mockAssetHoldingCacheHelper.cacheAssetHolding(address, emptyList()) }
+        coVerify(exactly = 1) { mockAccountInformationMapper(mockEntity, emptyAssetHoldings) }
```
mitsinsar: No need to verify these two, since you've already mocked these functions with these args

common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/domain/model/RegisteredAlgorandAccount.kt
```diff
@@ -0,0 +1,30 @@
...
+    val address: String,
+    val algoValue: String,
+    val usdValue: String,
+    val calculationType: String,
+    val accountExists: Boolean,
+    val isImportedToDB: Boolean,
+    val account: Int,
+    val change: Int,
+    val keyIndex: Int,
+    val derivationType: Int
```
mitsinsar: This model has so many responsibilities. Can you please separate those responsibilities into smaller parts

mitsinsar (changes requested):

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
```diff
@@ -29,32 +29,42 @@ internal class AddHdSeedUseCase @Inject constructor(
...
-            } ?: run {
-                return PeraResult.Error(Exception("Failed to insert hd seed"))
-            }
+    private suspend fun createNewSeed(entropy: ByteArray): PeraResult<Int> {
+        try {
+            val seed = peraBip39Sdk.getSeedFromEntropy(entropy)
+                ?: return PeraResult.Error(Exception("Failed to generate seed from entropy"))
+            val newSeedId = addHdSeed(seed, entropy)
+            setCustomInfo(newSeedId)
+            seed.fill(0)
```
yasin-ce, line 45: I guess this is for clearing cache but it's not easy to catch. Could you create a function for this? Which can be reused later.

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
```diff
@@ -29,32 +29,42 @@ internal class AddHdSeedUseCase @Inject constructor(
...
+            seed = seed,
+            entropy = entropy
+        ).toInt()
+    }
+
+    private suspend fun setCustomInfo(seedId: Int) {
+        customHdSeedInfoRepository.setCustomInfo(
+            CustomHdSeedInfo(
+                seedId = seedId,
+                entropyCustomName = "Wallet #$seedId",
```
yasin-ce, line 64: This magic string will cause issue when we support multilang.

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
```diff
@@ -29,32 +29,42 @@ internal class AddHdSeedUseCase @Inject constructor(
...
+            seed.fill(0)
+            return PeraResult.Success(newSeedId)
+        } catch (e: Exception) {
+            return PeraResult.Error(e)
         }
     }
+
+    private suspend fun addHdSeed(seed: ByteArray, entropy: ByteArray): Int {
+        return hdSeedRepository.addHdSeed(
+            seedId = 0, // ID will be auto-generated
```
yasin-ce, line 54: If it's auto-generated that would be better to remove this parameter to not create confusion.

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/model/AccountFastLookupResponse.kt
michaeltchuang: the API is returning string for this field.  The response parser should be string...but are you saying storing in db as double?  I could put this in mapper

Screenshot 2025-03-14 at 7 09 04 AM

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/model/AccountFastLookup.kt
michaeltchuang: what do you mean update their type?  have backend change the type their sending?  I think string is fine from API, and we can convert to `BigDecimal(stringValue)` on our end

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/model/AccountFastLookupResponse.kt
michaeltchuang: i think you're supposed to avoid `doubles` for money and use `BigDecimal` because of precision

common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/PeraBip39SdkImpl.kt
```diff
@@ -44,56 +59,119 @@ internal class PeraBip39SdkImpl @Inject constructor() : PeraBip39Sdk {
...
-        entropy = ByteArray(0) // delete secret key from memory
-        seed = ByteArray(0) // delete secret key from memory
+    override suspend fun fetchRegisteredAccounts(entropy: ByteArray): List<RegisteredAlgorandAccount> {
+        val output = mutableListOf<RegisteredAlgorandAccount>()
+        try {
+            val mnemonicCode = Mnemonics.MnemonicCode(entropy)
+            var seed = mnemonicCode.toSeed()
+            val xHDWalletAPI = XHDWalletAPIAndroid(seed)
+
+            for (accountIndex in 0 until 5) {
```
yasin-ce: You may create constant for the magic numbers. Also is this the  final logic or you will update it later? Because as I know you need to iterate 5 more after you found the last account.

yasin-ce (changes requested):

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/usecase/AccountInformationUseCases.kt
michaeltchuang: i'll change it for this, but the other `account/info` calls aren't returning `PeraResult`, so i was trying to follow convention you set of returning model in this module.  We should change rest in another PR to not add scope to my PR

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/domain/usecase/GetAccountDetailCacheStatusFlowUseCaseTest.kt
michaeltchuang: i'll just remove this test

common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/PeraBip39SdkImpl.kt
michaeltchuang: I'm looping 5 x 5 x 5 for now, but we can doublecheck next HD meeting.  I think let's merge what we have since its quite a big PR with just `common-sdk` changes, and then refactor the tiny extra changes later when we know final algorithm

common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/domain/model/RegisteredAlgorandAccount.kt
michaeltchuang: let me see if I can remove this file from PR since making so many changes in one PR makes it difficult

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
michaeltchuang: do you mean the `seed.fill(0)`?  that's just to clear the temporary seed variable in this function that Bruno would like us to start doing.  I'm not sure what you mean by needing a function

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
michaeltchuang: for the mapper, we sometimes use seedId...i think I prefer passing 0 for now (since that's one common industry convention), and we can refactor without it later after i land this big PR

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
michaeltchuang: totally understand...let's worry about this after account refactor and hd wallets goes out and we're going to figure how to inject android resource files in common-sdk.

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupFetchHelperImpl.kt
michaeltchuang: fixed

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupFetchHelper.kt
michaeltchuang: i renamed it to repository

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/data/repository/AccountInformationCacheHelperImplTest.kt
michaeltchuang: fixed

common-sdk/src/test/kotlin/com/algorand/wallet/account/info/data/mapper/model/AccountFastLookupMapperImplTest.kt
michaeltchuang: fixed

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/model/AccountFastLookup.kt
mitsinsar: Retrofit can do that for us. Also, domain model should have main data type which is BigInteger for algo value.

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
mitsinsar: I believe he suggested to create a function to make it understandable. For someone who doesn't know the context, it is not clear why we have this here. Something like below would be helpful;

`fun clearSeedFromMemory()...`

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/usecase/GetAccountFastLookupUseCase.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import com.algorand.wallet.account.info.data.mapper.model.AccountFastLookupMapper
+import com.algorand.wallet.account.info.data.repository.AccountFastLookupRepository
+import com.algorand.wallet.account.info.domain.model.AccountFastLookup
+import com.algorand.wallet.foundation.PeraResult
+import javax.inject.Inject
+
+internal class GetAccountFastLookupUseCase @Inject constructor(
+    private val accountFastLookupRepository: AccountFastLookupRepository,
+    private val accountFastLookupMapper: AccountFastLookupMapper
```
mitsinsar, line 23: I see that you moved this mapper into data layer but you still inject it here. You still depend on data layer. Domain layer shouldn't import anything from data layer.

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupRepository.kt
```diff
@@ -0,0 +1,20 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+
+package com.algorand.wallet.account.info.data.repository
+
+import com.algorand.wallet.account.info.data.model.AccountFastLookupResponse
+import com.algorand.wallet.foundation.PeraResult
+
+internal interface AccountFastLookupRepository {
```
mitsinsar, line 18: This interface should be in domain layer. Reason to have this interface is to separate layers and invert dependencies. If you put this interface into data, your domain depends on data. our architecture is domain driven, that's why this approach is wrong

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/usecase/GetAccountFastLookupUseCase.kt
```diff
@@ -0,0 +1,41 @@
...
+        try {
+            val response = accountFastLookupRepository.fetchAccountFastLookup(address)
+            response.getDataOrNull()?.let { data ->
+                val accountFastLookup = accountFastLookupMapper(
+                    response = data
+                )
+                return PeraResult.Success(accountFastLookup)
+            } ?: run {
+                throw IllegalArgumentException("Empty AccountFastLookupResponse object")
+            }
```
mitsinsar, line 36: Please don't use try catch. I explained the reason in previous pull requests.

mitsinsar (changes requested):

michaeltchuang:
converting this to draft since it's in a future build from now

michaeltchuang:
superseded by this PR - https://github.com/perawallet/pera-android/pull/232

PR #202, opened 2025-03-17
PERA-1469 :: Add account migration viewer

michaeltchuang:
# Pull Request Template

## Description
- Add migration viewer and related work

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1469

## Checklist

## Additional Notes
- I set orderIndex on migrate function
- I put this up earlier so anyone can use it.  I've randomly 
used to see accounts from pre v6 and post v6
- There is a feature flag that always shows the viewer on non `prodRelease` builds, and only if you have old accounts on `prodRelease`
- I think i want the payload change (so we don't have so many params in add account function) to be in a separate PR to keep PRs scope smaller
- I also need to fix `LedgerBle`, but maybe that can be a separate PR

Looks like this

Screenshot_1742348646


app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerFragment.kt
```diff
@@ -0,0 +1,285 @@
...
+            migrationViewerViewModel.viewEvent.collect { event ->
+                when (event) {
+                    is ViewEvent.NavigateBack -> {
+                        navBack()
+                    }
+                }
+            }
+        }
+
+        Column(
```
mitsinsar, line 99: Should we consider removing account lazy columns and make this scrollable? I haven't tested but this implementation may look weird when there are a lot of accounts. If you tested and it works properly, you can ignore this comment

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerMigrateUseCase.kt
```diff
@@ -0,0 +1,100 @@
...
+                    accountAdditionUseCase.addNewAccount(it)
+                    migratedCount++
+                }
+            }
+            return PeraResult.Success(migratedCount)
+        } catch (exception: Exception) {
+            return PeraResult.Error(exception)
+        }
+    }
+}
```
mitsinsar: I think we can create a private function that takes 3 args; localAccount, type and creation type. So that we can remove LongMethod warning

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerMigrateUseCase.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.algorand.android.models.Account
+import com.algorand.android.models.AccountCreation
+import com.algorand.android.usecase.AccountAdditionUseCase
+import com.algorand.android.usecase.GetLocalAccountsFromSharedPrefUseCase
+import com.algorand.android.utils.analytics.CreationType
+import com.algorand.wallet.encryption.domain.manager.AESPlatformManager
+import com.algorand.wallet.foundation.PeraResult
+import javax.inject.Inject
+
+class MigrationViewerMigrateUseCase @Inject constructor(
```
mitsinsar: Can we rename this use case? Since we are going to use this to for migration without using MigrationViewever, this naming doesn't seem right.

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerMigrateUseCase.kt
```diff
@@ -0,0 +1,100 @@
...
+                    else -> {}
+                }
+                migrateAccount?.let {
+                    accountAdditionUseCase.addNewAccount(it)
+                    migratedCount++
+                }
+            }
+            return PeraResult.Success(migratedCount)
+        } catch (exception: Exception) {
+            return PeraResult.Error(exception)
```
mitsinsar: Could you please log exception to firebase, if there is any, we can see the reason

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    sealed interface ViewState {
+        data object Loading : ViewState
+        data class Error(val error: String) : ViewState
```
mitsinsar, line 100: we basically show an error and navigate back, so this `error` is being used as `event`, not a `state`. 

Could you please more this error to ViewEvent? 

Also, don't forget to use `activityTag` for error notification, otherwise it disappears as soon as fragment dies.

app/src/test/kotlin/com/algorand/android/MigrationViewerMigrateUseCaseTest.kt
```diff
@@ -0,0 +1,160 @@
...
+        val secretKey = byteArrayOf(1, 2, 3, 4)
+        val encryptedSecretKey = byteArrayOf(5, 6, 7, 8)
+
+        val standardAccount = mockk<Account>()
+        every { standardAccount.address } returns "addr1"
+        every { standardAccount.name } returns "Account 1"
+        every { standardAccount.index } returns 0
+        every { standardAccount.isBackedUp } returns true
+        every { standardAccount.detail } returns mockk<Account.Detail.Standard>()
+        every { standardAccount.getSecretKey() } returns secretKey
```
mitsinsar: Instead of mocking, can you please create account object manually? There is no benefit of mocking if we are going to define every property of it.

Also creating manually would be safer for this test case.

To make function readable, can you create these accounts in `private companion object` as `val STANDARD_ACCOUNT`...

app/src/test/kotlin/com/algorand/android/MigrationViewerMigrateUseCaseTest.kt
```diff
@@ -0,0 +1,160 @@
...
+        every { aesPlatformManager.encryptByteArray(secretKey) } returns encryptedSecretKey
+
+        val result = sut.invoke()
+
+        assertTrue(result is PeraResult.Success)
+        assertEquals(4, (result as PeraResult.Success).data)
+
+        coVerify {
+            accountAdditionUseCase.addNewAccount(match {
+                it.address == "addr1" &&
```
mitsinsar: Please compare account object directly. Create the same object as expected (since we also encrypt before passing to usecase), then compare.

If I go to prod code, and change every field except address and its type, this function will still pass

common-sdk/src/main/kotlin/com/algorand/wallet/encryption/domain/manager/AESPlatformManagerImpl.kt
```diff
@@ -111,4 +111,8 @@ internal class AESPlatformManagerImpl @Inject constructor() : AESPlatformManager
 
         return String(plaintextBytes)
     }
+
+    override fun deleteSecretByteArrayFromMemory(): ByteArray {
```
mitsinsar: This function seems misleading. This doesn't delete it, it simply returns empty byte array.

mitsinsar (changes requested):

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerFragment.kt
michaeltchuang: `LazyColumn` should be fine, most of our users don't have that many accounts and may never even see this screen either

common-sdk/src/main/kotlin/com/algorand/wallet/encryption/domain/manager/AESPlatformManagerImpl.kt
michaeltchuang: i switch name to `clearSecretByteArrayFromMemory`...because setting it empty byte array once way to clear variable.  If you prefer a different name, just let me know and I can change it to that

app/src/test/kotlin/com/algorand/android/MigrationViewerMigrateUseCaseTest.kt
michaeltchuang: i switched to it to expected and asserted the two object properties...hopefully that's closer to what you wanted

app/src/test/kotlin/com/algorand/android/MigrationViewerMigrateUseCaseTest.kt
michaeltchuang: i created companion objects

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerViewModel.kt
michaeltchuang: i ended up just creating a placeholder error state screen.  What did you mean by activityTag?  the baseFragment references activity.  Now that it's a state and doesn't navBack unless user does it...maybe how i coded it is fine

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerMigrateUseCase.kt
michaeltchuang: i renamed it `MigrateTo6xUseCase`

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerMigrateUseCase.kt
michaeltchuang: ok fixed it to this way

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerMigrateUseCase.kt
michaeltchuang: fixed

i'll create a separate PR after account refactor is merged into `dev` branch to optimize other `firebaseCrashlytics` references to use `PeraEventTracker` logException

app/src/main/kotlin/com/algorand/android/ui/settings/developersettings/DeveloperSettingsViewModel.kt
```diff
@@ -38,4 +44,15 @@ class DeveloperSettingsViewModel @Inject constructor(
...
         }
     }
+
+    fun showMigrationViewer(): Boolean {
+        val isEnableAccountMigrationViewer = isFeatureToggleEnabled
+            .invoke(ENABLE_ACCOUNT_DB_MIGRATION_VIEWER)
+        if (isEnableAccountMigrationViewer) {
+            return true
+        } else {
+            return getLocalAccountsFromSharedPrefUseCase
```
mitsinsar: Based on this logic, even if we enable migration viewer, it will be visible when there are accounts. I think we should hide this button when toggle is disabled no matter what.

app/src/main/kotlin/com/algorand/android/ui/settings/migrationviewer/MigrationViewerViewModel.kt
mitsinsar: Yes, with placeholder, it became a state and this version is fine. 

Our `showGlobalError` function gets a tag. If tag is a current fragment's name, when you navigate back, error disappears as well. That's why, if we want to display the error even if we navigate back, we should use activity tag, instead of fragment tag

app/src/test/kotlin/com/algorand/android/MigrateTo6xUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+    private var peraEventTracker: PeraEventTracker = mockk()
+    private var sut: MigrateTo6xUseCase =
+        MigrateTo6xUseCase(
+            getLocalAccountsFromSharedPrefUseCase,
+            aesPlatformManager,
+            accountAdditionUseCase,
+            peraEventTracker
+        )
+
+    companion object {
```
mitsinsar: Let's make this companion object `private` and remove `private` modifiers in it.

Also, can you move this at the end of the file please, for consistency

common-sdk/src/main/kotlin/com/algorand/wallet/encryption/domain/manager/AESPlatformManagerImpl.kt
```diff
@@ -111,4 +111,8 @@ internal class AESPlatformManagerImpl @Inject constructor() : AESPlatformManager
 
         return String(plaintextBytes)
     }
+
+    override fun clearSecretByteArrayFromMemory(): ByteArray {
```
mitsinsar: Let's remove this function completely. This does nothing but returns empty byte array. You don't need `AESPlatformManager` to do that.

There are usages like `AlgoAccountSdkImpl` where you injected AESPlatformManager just to create a simple byte array. This creates unnecessary dependency.

mitsinsar (changes requested):

app/src/main/kotlin/com/algorand/android/ui/settings/developersettings/DeveloperSettingsViewModel.kt
michaeltchuang: the feature flag is only enabled currently on staging and debug builds so production users won't be able to re-migrate if they run into issues...but i think we can change the business logic after this PR gets merged if needed.  Maybe we see first if prod users even run into issues

app/src/test/kotlin/com/algorand/android/MigrateTo6xUseCaseTest.kt
michaeltchuang: updated

common-sdk/src/main/kotlin/com/algorand/wallet/encryption/domain/manager/AESPlatformManagerImpl.kt
michaeltchuang: i created an extension function in `encryption/domain/util` instead...would that be better?  I thought Yasin's feedback was setting byteArray to 0 and a comment beside it was not dev friendly enough so i tried to create a helper function.  

Or did you want to just go back to what it was?
`secretKey = ByteArray(0) // delete secret key from memory`

yasin-ce (approved):

mitsinsar (approved):

PR #210, opened 2025-03-19
PERA-1812 :: Add migrateTo6X check in app start

michaeltchuang:
# Pull Request Template

## Description
- Add migrateTo6X check in app start

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1812

## Checklist

## Additional Notes
- Migrate function PR should be merged first then we can merge this one afterwards


yasin-ce (approved):

app/src/main/kotlin/com/algorand/android/modules/settings/domain/model/SettingsPreview.kt
```diff
@@ -10,7 +10,7 @@
  * limitations under the License
  */
 
-package com.algorand.android.modules.settings.ui.model
+package com.algorand.android.modules.settings.domain.model
```
mitsinsar: This is a UI model, let's keep this in `ui` package

app/src/main/kotlin/com/algorand/android/modules/settings/domain/usecase/DeveloperSettingsPreviewUseCase.kt
```diff
@@ -10,7 +10,7 @@
  * limitations under the License
  */
 
-package com.algorand.android.modules.settings.ui.usecase
+package com.algorand.android.modules.settings.domain.usecase
```
mitsinsar: This is a use case for `ui/preview`, let's keep this in `ui` package. In the future, we won't have `...PreviewUseCase`  classses, things will be handled in VMs.

app/src/main/kotlin/com/algorand/android/modules/settings/domain/usecase/SettingsPreviewUseCase.kt
```diff
@@ -10,11 +10,11 @@
  * limitations under the License
  */
 
-package com.algorand.android.modules.settings.ui.usecase
+package com.algorand.android.modules.settings.domain.usecase
```
mitsinsar: Same, usecase for `ui`

app/src/main/kotlin/com/algorand/android/modules/settings/ui/mapper/SettingsPreviewMapper.kt
```diff
@@ -12,7 +12,7 @@
 
 package com.algorand.android.modules.settings.ui.mapper
 
-import com.algorand.android.modules.settings.ui.model.SettingsPreview
+import com.algorand.android.modules.settings.domain.model.SettingsPreview
```
mitsinsar: Same, lets keep this in `ui` package

app/src/main/kotlin/com/algorand/android/migration/MigrationManager.kt
```diff
@@ -18,10 +18,12 @@ import javax.inject.Singleton
...
 class MigrationManager @Inject constructor(
     private val accountMigrationManager: AccountMigrationManager,
-    private val encryptedPinMigrationManager: EncryptedPinMigrationManager
+    private val encryptedPinMigrationManager: EncryptedPinMigrationManager,
+    private val account6xMigrationManager: Account6xMigrationManager
 ) {
     fun makeMigrations() {
         encryptedPinMigrationManager.makeMigrationIfNeeded()
         accountMigrationManager.makeMigrationIfNeeded()
+        account6xMigrationManager.migrateTo6xIfNeeded()
```
mitsinsar: In the current implementation, migration is being done in main thread. That means, app is not being initialized until migration is completed.

Account6xMigrationManager uses custom coroutine scope, and when you call `migrateTo6xIfNeeded` it creates new coroutine scope and doesn't wait its job to finish. Which means, app may be initialized before migration is completed. We need to change our approach. 

`MigrationManager::makeMigrations` needs to be suspended, and we need to initialize everything (like, WalletConnect, LifecycleAwareComponents etc...) when migration is completed. If we don't do that, there is a high chance that we launch the app without waiting for migration to be completed.

mitsinsar (changes requested):

app/src/main/kotlin/com/algorand/android/modules/settings/domain/model/SettingsPreview.kt
michaeltchuang: we should move away from modules folder having anything ui...and put ui code in `com.algorand.android.ui` folder.  

module = data/di/domain folders for each section
ui = fragment/viewmodels for each section

app/src/main/kotlin/com/algorand/android/modules/settings/domain/usecase/DeveloperSettingsPreviewUseCase.kt
michaeltchuang: i can move it to a `com.algorand.android.ui.settings.usecase` for now but we should refactor this so this type of architecture is gone from project

app/src/main/kotlin/com/algorand/android/modules/settings/domain/usecase/SettingsPreviewUseCase.kt
michaeltchuang: moved to a ui folder

app/src/main/kotlin/com/algorand/android/migration/MigrationManager.kt
michaeltchuang: I added a suspend step in PeraApp so migration will happen first before walletconnect and other stuff will be initialized

app/src/main/kotlin/com/algorand/android/modules/settings/domain/utils/Constants.kt
```diff
@@ -0,0 +1,5 @@
+package com.algorand.android.modules.settings.domain.utils
+
+internal object Constants {
```
mitsinsar: This is so generic. Can you rename it to make it more clear please?

Every class can access this and no one will now the context. We shouldn't create an object/file named `constant` at all.

app/src/main/kotlin/com/algorand/android/core/PeraApp.kt
```diff
@@ -75,13 +79,18 @@ open class PeraApp : Application() {
...
         initializeFirebase()
         BaseViewModel.initialize(peraEventTracker)
-        migrationManager.makeMigrations()
         peraSecurityManager.initializeSecurityManager()
         AppCompatDelegate.setDefaultNightMode(sharedPref.getSavedThemePreference().convertToSystemAbbr())
 
-        initializeWalletConnect()
-        bindApplicationLifecycleAwareComponents()
-        bindActivityLifecycleAwareComponents()
+        CoroutineScope(Dispatchers.Main).launch {
```
mitsinsar: 1. This coroutine scope is not attached to any lifecycle
2. Account6xMigrationManager still creates completely independent coroutine, and parent coroutine will not wait it to be completed. So basically this will be the same as previous implementation. 
3. This class doesn't know how initialization functions like WC, Cache etc are implemented. If they don't use SupervisorJob for their coroutines, and if one of them throws an exception, all of them will be affected. That's why they shouldn't be in this scope

mitsinsar (changes requested):

app/src/main/kotlin/com/algorand/android/modules/settings/domain/utils/Constants.kt
michaeltchuang: i renamed it `SettingsConstants.kt`

app/src/main/kotlin/com/algorand/android/migration/MigrationManager.kt
michaeltchuang: @mitsinsar i modified `MigrationManager` to be more like `LifecycleAwareCacheManagerImpl`, please check if this is closer to what you wanted

PR #214, opened 2025-03-22
PERA-1540 :: Make ASB backward compatible

mitsinsar:
# Pull Request Template

## Description
- Moved restore backup function into common-sdk
- Remaining asb function will be moved after account refactor

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1540
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1795
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1829

## Checklist

https://github.com/user-attachments/assets/f5da5ede-56c6-4b53-8455-928970cbbdc4


common-sdk/src/main/kotlin/com/algorand/wallet/asb/domain/usecase/BackupProtocolConstants.kt
michaeltchuang, line 1: we've been putting `Constant.kt` file under `domain/util`...i think we should keep same convention?

common-sdk/src/test/kotlin/com/algorand/wallet/asb/domain/mapper/AsbBackupDataMapperImplTest.kt
```diff
@@ -0,0 +1,83 @@
...
+    private val sut = AsbBackupDataMapperImpl()
+
+    @Test
+    fun `EXPECT mapped asb backup payload`() {
+        val result = sut(BACKUP_PROTOCOL_PAYLOAD)
+
+        val expected = AsbBackupData(
+            deviceId = "device_id",
+            providerName = "provider_name",
+            accounts = listOf(WATCH_ACCOUNT, STANDARD_ACCOUNT)
```
michaeltchuang: since this is going into `common-sdk` and not app module, should we name these variables to like `ALGO25_ACCOUNT` or something like that to align with our new account type naming convention?  What happens when we add HD wallet code...does standard mean we have secret (regardless of algo25 or bip39)?

yasin-ce (approved):

common-sdk/src/main/kotlin/com/algorand/wallet/asb/domain/usecase/BackupProtocolConstants.kt
mitsinsar: I moved into `domain/util` folder, however, since this file is public, I prefer keeping it as `BackupProtocolConstant` so that classes that uses this variable know where it comes from. Otherwise it looks like `Constants.SINGLE_ACCOUNT_TYPE_NAME` and this doesn't give any context.

app/src/main/kotlin/com/algorand/android/modules/asb/importbackup/accountselection/utils/AsbAccountImportParser.kt
```diff
@@ -75,12 +75,7 @@ class AsbAccountImportParser @Inject constructor(
         return isAccountTypeEligible
     }
 
-    private fun isAccountTypeEligible(accountTypeName: String?): Boolean {
-        val accountType = Account.Type.valueOf(accountTypeName ?: return false)
-        return AlgorandSecureBackupUtils.eligibleAccountTypes.contains(accountType)
-    }
-
     private fun isStandardAccount(accountTypeName: String?): Boolean {
```
michaeltchuang: could we change this function to `isAlgo25Account` as well?

michaeltchuang (approved):
doesn't crash on my test ASB txt problem file when i import so this is already an improvement.  

I had a function name change i think might make sense before merging, but not a blocker for this PR

app/src/main/kotlin/com/algorand/android/modules/asb/importbackup/accountselection/utils/AsbAccountImportParser.kt
mitsinsar: Oh nice catch! Will do that 👍

PR #225, opened 2025-03-28
PERA-1884 :: Fix backup flow crash on Account Detail

michaeltchuang:
# Pull Request Template

## Description
- Fix backup flow crash on Account Detail
- Renamed publicKeyofAccountToBackup to accountToBackup because are working with addresses and not public keys (so this PR touched more files)

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1884
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1800

## Checklist

## Additional Notes
- Also fixed the updateState bug since UI wasn't updating otherwise


app/src/main/kotlin/com/algorand/android/ui/register/PassphraseValidationViewModel.kt
```diff
@@ -33,11 +33,16 @@ class PassphraseValidationViewModel @Inject constructor(
...
     private val accountManager: AccountManager,
     private val aesPlatformManager: AESPlatformManager,
+    private val algoAccountSdk: AlgoAccountSdk,
     private val getAlgo25SecretKey: GetAlgo25SecretKey,
     private val peraBip39Sdk: PeraBip39Sdk,
     private val stateDelegate: StateDelegate<ViewState>
 ) : BaseViewModel(), StateViewModel<PassphraseValidationViewModel.ViewState> by stateDelegate {
 
+    init {
+        stateDelegate.setDefaultState(ViewState.Loading)
```
mitsinsar: Can we add Idle state for default state please? In the initialization process, nothing happens and we "ASSUME" that it is being loaded. Loading must be set when we call "do this", "do that". Adding another state seems extra work, but it prevents us opening VM and checking it to see what is going on there. 

When we write tests, there may be cases where we don't need loading. Because functions need to be tested separately. Please add Idle for default states, this will make things more predictable, testable and aligns with unidirectional data flow principle.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/ui/AccountDetailFragment.kt
```diff
@@ -454,7 +455,13 @@ class AccountDetailFragment :
         nav(
             AccountDetailFragmentDirections
                 .actionAccountDetailFragmentToBackupPassphraseInfoNavigation(
-                    publicKeysOfAccountsToBackup = arrayOf(accountDetailViewModel.accountAddress)
+                    accountsToBackup = arrayOf(accountDetailViewModel.accountAddress),
+                    onboardingAccountType = if (accountDetailViewModel.accountType ==
+                        com.algorand.wallet.account.detail.domain.model.AccountType.HdKey) {
```
mitsinsar: Let's fix the import. Using like "AccountType.HdKey" should be fine and it shouldn't conflict with OnboardingAccountType

mitsinsar (changes requested):

app/src/main/kotlin/com/algorand/android/ui/register/PassphraseValidationViewModel.kt
michaeltchuang: i'm ok with adding an `Idle` state but then we need to not show a blank screen for any successful state (we can start fixing this if screen is implemented in compose, since that's a terrible user experience otherwise)...something like "Content will appear shortly" for `Idle` UI state at minimum should suffice

app/src/main/kotlin/com/algorand/android/modules/accountdetail/ui/AccountDetailFragment.kt
michaeltchuang: nice catch, fixed

app/src/main/kotlin/com/algorand/android/ui/register/PassphraseValidationViewModel.kt
michaeltchuang: I updated current file to use `Idle` instead of `Loading`

mitsinsar: There won't be blank screen since we call "initialize" in on create. It will trigger loading immediately. This won't affect user experience. This is about clean code.

mitsinsar (approved):

PR #232, opened 2025-04-01
PERA-1581 :: Add account fast lookup network calls

michaeltchuang:
# Pull Request Template

## Description
- Add account fast lookup code into common sdk (used in hd recovery account flow)

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1581

## Checklist

## Additional Notes
- This is just the common-sdk code so UI code PR for account scanning will be much smaller
- If there is time to review this during `v6.0.0` and it can be included, great...if not, it can also go in next build


common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/model/AccountFastLookupResponse.kt
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+
+package com.algorand.wallet.account.info.data.model
+
+import com.google.gson.annotations.SerializedName
+
+internal data class AccountFastLookupResponse(
```
mitsinsar, line 17: Let's make all the fields nullable. We wouldn't want to get failure just because one of the fields we get is null

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/model/AccountFastLookup.kt
```diff
@@ -0,0 +1,10 @@
+package com.algorand.wallet.account.info.domain.model
+
+import java.math.BigDecimal
+
+data class AccountFastLookup(
+    val algoValue: BigDecimal,
+    val usdValue: BigDecimal,
+    val calculationType: String,
```
mitsinsar: Since this is a domain model, either we should use enum for calculation type (MIN, EXACT), or we should delete this field. If somehow we use `calculationType` as string in domain model, it would create an issue in the future. 

I think we can simply remove this field since we don't use it right now.

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupRepositoryImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+    ): PeraResult<AccountFastLookup> {
+        return request { api.getAccountFastLookup(address) }.use(
+            onSuccess = { response ->
+                val accountFastLookup = accountFastLookupMapper.invoke(response = response)
+                PeraResult.Success(accountFastLookup)
+            },
+            onFailed = { exception, code ->
+                PeraResult.Error(exception, code)
+            }
+        )
```
mitsinsar: I think instead of doing this, you can simply do this;

```
request { api.getAccountFastLookup(address) }.map {
    accountFastLookupMapper(response)
}
```

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupRepositoryImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+internal class AccountFastLookupRepositoryImpl @Inject constructor(
+    private val api: AccountFastLookupApiService,
+    private val accountFastLookupMapper: AccountFastLookupMapper
+) : AccountFastLookupRepository {
+
+    override suspend fun fetchAccountFastLookup(accountAddress: String): PeraResult<AccountFastLookup> {
+        return fetch(accountAddress)
+    }
+
+    private suspend fun fetch(
```
mitsinsar: Either we should rename this, or remove this function. If we add another function in this class in the future, this naming would force us to rename it.

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
```diff
@@ -26,38 +25,47 @@ internal class AddHdSeedUseCase @Inject constructor(
...
-                    )
-                )
-                seed.clearFromMemory()
-                return PeraResult.Success(newSeedIdInDB)
-            } ?: run {
-                return PeraResult.Error(Exception("Failed to insert hd seed"))
-            }
+    private suspend fun createNewSeed(entropy: ByteArray): PeraResult<Int> {
+        try {
+            val seed = peraBip39Sdk.getSeedFromEntropy(entropy)
```
mitsinsar: Let's not use `try-catch` here. sdk function is already wrapped with try&catch.

mitsinsar (changes requested):

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/domain/model/AccountFastLookup.kt
michaeltchuang: i removed the field

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/model/AccountFastLookupResponse.kt
michaeltchuang: i made them nullable

common-sdk/src/main/kotlin/com/algorand/wallet/account/core/domain/usecase/AddHdSeedUseCase.kt
michaeltchuang: i removed it

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupRepositoryImpl.kt
michaeltchuang: i switched it to this

common-sdk/src/main/kotlin/com/algorand/wallet/account/info/data/repository/AccountFastLookupRepositoryImpl.kt
michaeltchuang: i removed this

mitsinsar (approved):

PR #236, opened 2025-04-02
PERA-1800 :: Set accountCreation to be null during backup process

michaeltchuang:
# Pull Request Template

## Description
- We need to set `accountCreation` to be null during backup process
- Fix backup banner for hd wallet too
- Hide warning on account icon if amount is 0 like backup banner

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1800

## Checklist

## Additional Notes
- Easiest way to test is create new account on testnet, send some money to it so it's non-zero...and backup banner should show on account detail page


app/src/main/kotlin/com/algorand/android/ui/register/BackupPassphraseViewModel.kt
```diff
@@ -48,37 +54,73 @@ class BackupPassphraseViewModel @Inject constructor(
...
-    fun getMnemonic(args: BackupPassphraseFragmentArgs) {
-        viewModelScope.launchIO {
-            val encryptedEntropy = (args.accountCreation?.type as? AccountCreation.Type.HdKey)?.encryptedEntropy
-            val passphrase = encryptedEntropy?.let {
-                val entropy = aesPlatformManager.decryptByteArray(it)
-                peraBip39Sdk.getMnemonicFromEntropy(entropy)
-            } ?: run {
-                val encryptedAlgo25Key =
-                    (args.accountCreation?.type as? AccountCreation.Type.Algo25)?.encryptedSecretKey
+    suspend fun getMnemonic(args: BackupPassphraseFragmentArgs): List<String> {
```
mitsinsar: Please remove `suspend` modifier. We are calling this function from fragment. If operation needs to be suspended, scope needs to be created by using `viewModelScope`. 

Also returned list is not being used

mitsinsar (changes requested):
Backup and creation needs to be separated into two different components. Could you please create a ticket to refactor these flows? We shouldn't combine creation and backup flows. We can merge after fixing the comment, and refactor with different ticket

michaeltchuang:
> Backup and creation needs to be separated into two different components. Could you please create a ticket to refactor these flows? We shouldn't combine creation and backup flows. We can merge after fixing the comment, and refactor with different ticket

I also agree, I was just trying to fix the current architecture.  I also wanted to add the idea we should move the viewmodels for onboarding (and maybe backup too) when we had time to `common-sdk` like `ViewPassphraseViewModel.kt` since i think they long term should belong there

Here's jira ticket in backlog - https://algorandfoundation.atlassian.net/browse/PERA-1930

app/src/main/kotlin/com/algorand/android/ui/register/BackupPassphraseViewModel.kt
michaeltchuang: Fixed

yasin-ce (approved):

PR #245, opened 2025-04-08
PERA-1939 :: AccountStatusDetail conversion to event/state structure

michaeltchuang:
# Pull Request Template

## Description
- AccountStatusDetail conversion to event/state structure (refactor out legacy files like ui usecase)
- Fix Learn more urls in AccountStatusDetailBottomSheet
- Rename Bip39 to One Key (it will likely to change to Universal but we will open a PR for copy change when we get the new hd wallet name)

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1939
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1934

## Checklist

## Additional Notes
- Please test account status detail bottom sheet works as expected still (click top right wallet icon in account detail screen)


michaeltchuang:
recording of it

[accountstatusdetail.webm](https://github.com/user-attachments/assets/cce14e16-2909-40c2-9c84-bc66dc09d41c)

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -15,53 +15,155 @@ package com.algorand.android.modules.accountdetail.accountstatusdetail.ui
...
+) : BaseViewModel(), StateViewModel<ViewState> by stateDelegate, EventViewModel<ViewEvent> by eventDelegate {
 
     private val navArgs = AccountStatusDetailBottomSheetArgs.fromSavedStateHandle(savedStateHandle)
     val accountAddress = navArgs.accountAddress
-    val authAccountAddress: String
-        get() = _accountStatusDetailPreviewFlow.value?.authAccountDisplayName?.accountAddress.orEmpty()
 
-    private val _accountStatusDetailPreviewFlow = MutableStateFlow<AccountStatusDetailPreview?>(null)
-    val accountStatusDetailPreviewFlow: StateFlow<AccountStatusDetailPreview?> get() = _accountStatusDetailPreviewFlow
+    val authAccountAddress: String
```
mitsinsar: Seems like this is not being used

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -15,53 +15,155 @@ package com.algorand.android.modules.accountdetail.accountstatusdetail.ui
...
+                currentState.authAccountDisplayName?.accountAddress.orEmpty()
+            } else {
+                ""
+            }
+        }
 
     init {
-        initAccountStatusDetailPreview()
+        stateDelegate.setDefaultState(ViewState.Idle)
+        loadAccountStatusDetail()
```
mitsinsar: Calling something like `loadAccountStatusDetail()` inside `init` isn’t ideal. That kind of work should be triggered explicitly by the Fragment/Dialog based on its own lifecycle events.

The init block should be reserved for basic configuration like setting default state values (e.g., stateDelegate.setDefaultState(ViewState.Idle)). When the ViewModel starts doing work like loading data during initialization, it becomes harder to control and predict — especially when testing.

If there is a function unrelated to state of this ViewModel, when we want to write its test, we must mock everything, even if they are not related with the function we want to test. 

I’ve mentioned this a few times already, and I don’t want to sound pushy — but I’d really appreciate if we could stick to this pattern. If you’re still not convinced, we can absolutely sit down and discuss it together and decide on a standard.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -15,53 +15,155 @@ package com.algorand.android.modules.accountdetail.accountstatusdetail.ui
...
 
-    private fun initAccountStatusDetailPreview() {
-        viewModelScope.launchIO {
-            accountStatusDetailPreviewUseCase.getAccountStatusDetailPreviewFlow(
-                accountAddress = accountAddress
-            ).collectLatest { preview ->
-                _accountStatusDetailPreviewFlow.emit(preview)
+    private fun loadAccountStatusDetail() {
+        viewModelScope.launch {
+            stateDelegate.updateState { ViewState.Loading }
```
mitsinsar: Please move this state update out of `viewModelScope.launch` scope. 

We shouldn't update states like this because `updateState` is executed asynchronously, which means updating state doesn't happen immediately. It usually doesn't cause any trouble on production, however in test functions - like `runTest` - it may cause to have flaky tests.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -15,53 +15,155 @@ package com.algorand.android.modules.accountdetail.accountstatusdetail.ui
...
+        data class Content(
+            val titleString: String? = null,
+            val accountOriginalTypeDisplayName: AccountDisplayName? = null,
+            val accountOriginalTypeIconDrawablePreview: AccountIconDrawablePreview? = null,
+            val accountOriginalActionButton: AccountAssetItemButtonState? = null,
+            val authAccountDisplayName: AccountDisplayName? = null,
+            val authAccountIconDrawablePreview: AccountIconDrawablePreview? = null,
+            val authAccountActionButton: AccountAssetItemButtonState? = null,
+            val accountTypeDrawablePreview: AccountIconDrawablePreview? = null,
+            val accountDetail: AccountDetail? = null,
```
mitsinsar: Instead of keeping account detail, you can pass new class called `DescriptionDetail` 

```
data class Content(
    val descriptionDetail: DescriptionDetail,
    ...
) {
    data class DescriptionDetail(
        val annotatedString: AnnotatedString,
        val hyperlinkUrl: String
    )
}
```

And then you can change `accountStatusDetailPreviewDecider.decideDescriptionAnnotatedString` return type from `AnnotatedString` to `DescriptionDetail` 

Because there is no reason to keep this login in bottom sheet, especially description and hyperlink are closely coupled.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -15,53 +15,155 @@ package com.algorand.android.modules.accountdetail.accountstatusdetail.ui
...
+            val accountTypeString: String? = null,
+            val descriptionAnnotatedString: AnnotatedString? = null,
+            val isRekeyGroupVisible: Boolean? = null,
+            val isRekeyToLedgerAccountVisible: Boolean? = null,
+            val isRekeyToStandardAccountVisible: Boolean? = null
+        ) : ViewState
```
mitsinsar: This kind of approach becomes messy when we add more types. Consider adding HdKey. Things get more complicated and this class gets bigger and bigger. 

The best solution would be to create separate bottom sheets for different account types. If we don't want to do that right now, other option would be updating state to something like this;

```
        data class Content(
            val accountOriginalTypeDisplayName: AccountDisplayName? = null,
            val accountOriginalTypeIconDrawablePreview: AccountIconDrawablePreview? = null,
            val accountOriginalActionButton: AccountAssetItemButtonState? = AccountAssetItemButtonState.COPY,
            val accountTypeDrawablePreview: AccountIconDrawablePreview? = null,
            val accountTypeString: String? = null,
            val descriptionAnnotatedString: AnnotatedString? = null,
            val type: ContentAccountType
        ) {

            sealed interface ContentAccountType {
                data object Algo25 : ContentAccountType
                data object Ledger : ContentAccountType
                data object NoAuth : ContentAccountType
                data object HdKey : ContentAccountType
                data object Rekeyed : ContentAccountType
                data object RekeyedAuth : ContentAccountType
            }
        }
```

By doing that, you wouldn't have to create `Booleans` in the states, which are hard to manage. For example you can display RekeyToLedgerAccount buttons if ContentAccountType is not Rekeyed or NoAuth, and this decision can be made in BottomSheet. If you want to keep this logic in ViewModel and make it testable, then you can create flags like; 

```
       sealed interface ContentAccountType {

                val isRekeyGroupVisible: Boolean 
                
                data object Algo25 : ContentAccountType {
                    override val isRekeyGroupVisible: Boolean
                        get() = false
                }
                
                data object Rekeyed : ContentAccountType {
                    override val isRekeyGroupVisible: Boolean
                        get() = true
                }

                ...
            }
```

Which prevents settings flags separately. However, I don't think this is necessary right now. 

So basically, keep common things in the `Content` class, and add specific things into sub classes if necessary.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -15,53 +15,155 @@ package com.algorand.android.modules.accountdetail.accountstatusdetail.ui
...
+        eventDelegate.sendEvent(viewModelScope, ViewEvent.CopyAccountAddressToClipboard(accountAddress))
+    }
+
+    fun onRekeyToStandardAccountClicked() {
+        eventDelegate.sendEvent(viewModelScope, ViewEvent.NavigateToRekeyToStandardAccount)
+    }
+
+    fun onRekeyToLedgerAccountClicked() {
+        eventDelegate.sendEvent(viewModelScope, ViewEvent.NavigateToRekeyToLedgerAccount)
+    }
```
mitsinsar: We can remove these functions, and call navigation or other functions directly in fragment. Since they don't have any logic, or they don't update the state, there is no reason to create these functions.

Additionally, `onSomethingHappened` kind of naming is more appropriate for UI layer code (like in a Fragment or Compose event handler), where you’re describing what happened.

In ViewModels, it’s better to name functions based on the action being performed, not the event that triggered it. So instead of describing what happened, describe what the ViewModel should do. 

These functions will be removed, because they are not really doing anything, so it is hard to explain this with these examples. But if we want to -let's say-  add an account after a button click, we shouldn't name it like `onAddAccountButtonClick`, we should name it `addAccount` instead.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
```diff
@@ -18,246 +18,208 @@ import android.view.View
...
 
-    private val accountTypeDrawablePreviewCollector: suspend (
-        AccountIconDrawablePreview?
-    ) -> Unit = { drawablePreview ->
-        if (drawablePreview != null) {
-            binding.accountStateTextView.apply {
-                val drawable = AccountIconDrawable.create(context, R.dimen.spacing_xxxxlarge, drawablePreview)
-                setDrawable(start = drawable)
+    private fun initObservers() {
+        lifecycleScope.launch {
```
mitsinsar: Please use `collectLatestOnLifecycle` instead of this.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
```diff
@@ -18,246 +18,208 @@ import android.view.View
...
-    }
 
-    private val accountOriginalTypeIconDrawablePreviewCollector: suspend (
-        AccountIconDrawablePreview?
-    ) -> Unit = { drawablePreview ->
-        if (drawablePreview != null) {
-            binding.accountItemView.apply {
-                val drawable = AccountIconDrawable.create(context, R.dimen.spacing_xxxxlarge, drawablePreview)
-                setStartIconDrawable(drawable)
+        lifecycleScope.launch {
```
mitsinsar: Please use `collectLatestOnLifecycle` instead of this.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
```diff
@@ -18,246 +18,208 @@ import android.view.View
...
-            setTitleText(displayName?.primaryDisplayName)
-            setDescriptionText(displayName?.secondaryDisplayName)
-            setOnLongClickListener { onAccountAddressCopied(displayName?.accountAddress.orEmpty()); true }
-        }
-    }
+            state.accountOriginalTypeIconDrawablePreview?.let { drawablePreview ->
+                accountItemView.apply {
+                    val drawable = AccountIconDrawable.create(context, R.dimen.spacing_xxxxlarge, drawablePreview)
+                    setStartIconDrawable(drawable)
+                }
```
mitsinsar: When you want to set/update multiple fields, it is ok to use `apply`. However, if you want to call a single function, using `apply` creates unnecessary lines and indentation. So instead of doing this, you can simply do;

``` 
val drawable = AccountIconDrawable.create(binding.root.context, R.dimen.spacing_xxxxlarge, drawablePreview)
accountItemView.setStartIconDrawable(drawable)
```

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
```diff
@@ -18,246 +18,208 @@ import android.view.View
...
-        event?.consume()?.run { onAccountAddressCopied(accountStatusDetailViewModel.accountAddress) }
-    }
+            accountTypeTextView.text = state.titleString
+            accountStateTextView.text = state.accountTypeString
 
-    private val navToUndoRekeyNavigationEventCollector: suspend (Event<Unit>?) -> Unit = { event ->
-        event?.consume()?.run { navToUndoRekeyNavigation() }
-    }
+            state.accountTypeDrawablePreview?.let { drawablePreview ->
+                accountStateTextView.apply {
```
mitsinsar: Same as above

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
```diff
@@ -18,246 +18,208 @@ import android.view.View
...
-        }
-    }
 
-    private val accountOriginalActionButtonCollector: suspend (AccountAssetItemButtonState?) -> Unit = { buttonState ->
-        if (buttonState != null) {
-            binding.accountItemView.apply {
-                setButtonState(buttonState)
-                setActionButtonClickListener { accountStatusDetailViewModel.onAccountActionButtonClicked() }
+            state.authAccountIconDrawablePreview?.let { drawablePreview ->
+                authAccountItemView.apply {
```
mitsinsar: Same as above

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
```diff
@@ -18,246 +18,208 @@ import android.view.View
...
 
-    private val authAccountDisplayNameCollector: suspend (AccountDisplayName?) -> Unit = { displayName ->
-        binding.authAccountItemView.apply {
-            setTitleText(displayName?.primaryDisplayName)
-            setDescriptionText(displayName?.secondaryDisplayName)
-            setOnLongClickListener { onAccountAddressCopied(displayName?.accountAddress.orEmpty()); true }
-        }
-    }
+    @Suppress("LongMethod")
+    private fun renderContentState(state: ViewState.Content) {
```
mitsinsar, line 93: This function is really long. Can you please make it smaller by extracting things into different functions.

mitsinsar (changes requested):

michaeltchuang:
i just noticed this on my PR branch (unknown) that I'll fix too

Screenshot_1744397973

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: I prefer to the bare minimum right now, and when Yasin has time...to refactor this to compose and make all these potential changes.  If things work in the PR, I'd just roll with what we have for now, and work in phases

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
michaeltchuang: changed

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: changed

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: changed

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
michaeltchuang: fixed

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: i moved it out of viewmodel scope

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: maybe we can talk about this Monday, because this seems like the last potential issue before landing this PR

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: removed, good catch

mitsinsar (approved):

PR #257, opened 2025-04-17
PERA-1949 :: Fix ARC-59 Crash

mitsinsar:
# Pull Request Template

## Description
When user doesn't have enough algo balance to cover fees, Sdk was throwing an exception. Added a balance check before creating transactions.

## Related Issues
Closes https://algorandfoundation.atlassian.net/browse/PERA-1949

## Checklist

# Case which causes the crash
Flow below was throwing an exception. Added an error and disabled button when sender doesn't have enough balance to cover Arc59 fees. 
https://github.com/user-attachments/assets/2b4a738a-2ec4-439c-8985-0599de0f5972

# Regular asset transactions
https://github.com/user-attachments/assets/ad6f0f54-4cf8-4989-b6ee-e2ba10bd2dfb


app/src/main/kotlin/com/algorand/android/ui/send/transferpreview/AssetTransferPreviewViewModel.kt
```diff
@@ -135,35 +143,61 @@ class AssetTransferPreviewViewModel @Inject constructor(
...
+    }
+
+    @Suppress("MagicNumber")
+    private suspend fun getTransactionFee(transactionData: TransactionSignData.Send): Long {
+        /**
+         * To be able to calculate transaction fees, we need to create transaction.
+         * To be able to create transactions, we need to make sure that the sender has enough balance.
+         * Otherwise SDK throws exception. This transaction is being used to calculate the fee only, not to send.
+         */
+        val dummyTransactionData = transactionData.copy(senderAlgoAmount = BigInteger.valueOf(999_999_999L))
```
michaeltchuang, line 174: i think there is a simulate function we can use in go algosdk version (to figure out txn fee), maybe we can look into that after hd wallet is out

i created a ticket for it - https://algorandfoundation.atlassian.net/browse/PERA-2020

yasin-ce (approved):

PR #287, opened 2025-05-09
PERA-2145 :: Remove try-catch from AccountInformationRepository

mitsinsar:
## Description
Reason of the crash was ConcurrentHashMap. Even though we set value nullable `ConcurrentHashMap<Key, Value?>`, it does not accept nullable values due to concurrency -check [here](https://www.javapedia.net/Concurrent-collections/1485) for the reason- . 

When we use setter below, Kotlin uses the setter in `Maps.kt`, which is generic and let us set null since our `value` is nullable. That's why IDE didn't warn at all. 
```
val map = ConcurrentHashMap<String, String?>()
map[""] = null
```

Maps.kt - setter
```
fun <K, V> MutableMap<K, V>.set(key: K, value: V)
```

If we need to use `ConcurrentHashMap` again, we need to make sure that we are using `map.put(K, V)` function directly instead of `map[K] = V`

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2145

## Checklist


michaeltchuang (approved):
doesn't seem to crash for me with this change, but we can see firebase logs later after more people play with it

PR #289, opened 2025-05-10
PERA-2157 :: Call initialize when lifecycle state is created

mitsinsar:
## Description
We collect when the lifecycle state is CREATED, but sometimes event is being triggered right before lifecycle state is set to CREATED, that's why it emits event but activity doesn't collect it.

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2157

## Checklist


michaeltchuang (approved):
working now video

[toggle_working.webm](https://github.com/user-attachments/assets/e228e53a-cc70-49ef-96be-b73b17e43e75)

PR #295, opened 2025-05-15
PERA-1962 :: Fix cards arbitrary signing for hd accounts

michaeltchuang:
# Pull Request Template

## Description
- Fix cards arbitrary signing for hd accounts

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1962

## Checklist

## Additional Notes
- Please doublecheck if you can arbitrary sign on cards
- I added `net.pwall.json:json-kotlin-schema` library since we'll need it when for liquid auth later

Video
[hd_arb_sign.webm](https://github.com/user-attachments/assets/d215f5db-596a-46b5-ac00-e178f03e832a)


common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt
```diff
@@ -48,7 +51,45 @@ internal class SignHdKeyTransactionImpl @Inject constructor() : SignHdKeyTransac
...
+            val prefixedData = prefixData(transactionByteArray)
+
+            val xHDWalletAPI = XHDWalletAPIAndroid(seed)
+            val (accountIndex, changeIndex, keyIndex) = listOf(
+                account.toUInt(),
+                change.toUInt(),
+                key.toUInt()
+            )
+
+            val stx = xHDWalletAPI.signAlgoTransaction(
```
yasin-ce: Here I guess we should use the function I shared below according to the description. Is there something I'm missing? 
```  /**
   * Sign arbitrary but non-Algorand related data
   * @param context
   * - context of the key (i.e Address, Identity)
   * @param account
   * - account number. This value will be hardened as part of BIP44
   * @param keyIndex
   * - key index. This value will be a SOFT derivation as part of BIP44.
   * @param data
   * - data to be signed in raw bytes
   * @param metadata
   * - metadata object that describes how `data` was encoded and what schema to use to validate
   * against
   *
   * @returns
   * - signature holding R and S, totally 64 bytes
   */
  fun signData(
          context: KeyContext,
          account: UInt,
          change: UInt,
          keyIndex: UInt,
          data: ByteArray,
          metadata: SignMetadata,
          derivationType: Bip32DerivationType = Bip32DerivationType.Peikert
  ): ByteArray {

    val valid = validateData(data, metadata)

    if (!valid) { // failed schema validation
      throw DataValidationException("Data validation failed")
    }

    return rawSign(
            getBIP44PathFromContext(context, account, change, keyIndex),
            data,
            derivationType
    )
  }```

**Review by yasin-ce — CHANGES_REQUESTED** [review]

File: common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt (hunk 1 above)
Comment (michaeltchuang) [reply common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt]: Good to doublecheck...I did try the `signData` approach but I had issues when it came to validation of schema (metadata variable) when using it...maybe we're not following https://github.com/algorandfoundation/xHD-Wallet-API-kt/blob/main/XHDWalletAPI-JVM/src/test/resources/msg.schema.json for cards?  do we need another schema file perhaps?  

iOS got it working so i just copied their approach - https://github.com/perawallet/perawallet-ios/pull/1567 (which is what this PR is currently)

I was gonna doublecheck with @PhearZero today if we could create a schema file that matched iOS's working approach since I would prefer using `signData` as well

Comment (michaeltchuang) [reply common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt]: @PhearZero also dependency `net.pwall.json:json-kotlin-schema` has a newer version that i think we might need to update [xHD-Wallet-API-kt](https://github.com/algorandfoundation/xHD-Wallet-API-kt) so `validateData` will still work (i got an error when i used a newer version on my end)

Comment (michaeltchuang) [reply common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt]: I talked to @PhearZero and we settled on `rawSign` instead of iOS approach for now.  Basically there is not a way to validate schema with prefix and we're not sure [xHD-Wallet-API-kt](https://github.com/algorandfoundation/xHD-Wallet-API-kt) will ever support this.

File: common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt [hunk 2]
```diff
@@ -48,7 +53,47 @@ internal class SignHdKeyTransactionImpl @Inject constructor() : SignHdKeyTransac
...
+            )
+
+            return stx
+        } catch (e: Exception) {
+            peraExceptionLogger.logException(e)
+            null
+        }
+    }
+
+    private fun prefixData(data: ByteArray): ByteArray {
```
yasin-ce, line 100: Having a small comment here can be helpful for later about why we use "MX" prefix.

common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt
```diff
@@ -48,7 +53,47 @@ internal class SignHdKeyTransactionImpl @Inject constructor() : SignHdKeyTransac
...
+            val prefixedData = prefixData(transactionByteArray)
+
+            val xHDWalletAPI = XHDWalletAPIAndroid(seed)
+            val (accountIndex, changeIndex, keyIndex) = listOf(
+                account.toUInt(),
+                change.toUInt(),
+                key.toUInt()
+            )
+
+            val stx = xHDWalletAPI.rawSign(
```
yasin-ce, line 83: Please add a comment about why we used `rawSign` instead of `signData`,

gradle/libs.versions.toml
```diff
@@ -141,6 +142,7 @@ navigation-ui-ktx = { module = "androidx.navigation:navigation-ui-ktx", version.
 
 # Networking / Json Parsing
 converter-gson = { module = "com.squareup.retrofit2:converter-gson", version.ref = "converterGson" }
+json-kotlin-schema = { module = "net.pwall.json:json-kotlin-schema", version.ref = "jsonKotlinSchema" }
```
yasin-ce: Do we need this library?

yasin-ce (changes requested):

gradle/libs.versions.toml
michaeltchuang: i was adding it since we will need it for liquid auth later but we can add it in that PR...i removed it for now

common-sdk/src/main/kotlin/com/algorand/wallet/algosdk/transaction/sdk/SignHdKeyTransactionImpl.kt
michaeltchuang: i added the comment

yasin-ce (approved):

PR #300, opened 2025-05-17
PERA-2222 :: Add custom wallet name to account information

michaeltchuang:
# Pull Request Template

## Description
- Add custom wallet name to account information

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2222

## Checklist

## Additional Notes
- We need to land this first so account status detail bottomsheet hd widget is easier to land
- I switched account status detail back to use account detail / information instead of account lite because loading all the parent hd wallet info into account lite isn't needed
- `navToHdScanNewAddresses` will be used in hd widget, I'm just trying to merge data layer changes first

### DIFF
--- a/app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
+++ b/app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
@@ -82,6 +82,9 @@ class AccountStatusDetailBottomSheet : BaseBottomSheet(R.layout.bottom_sheet_acc
             is ViewEvent.NavigateToRekeyToLedgerAccount -> {
                 navToRekeyToLedgerAccountNavigation()
             }
+            is ViewEvent.NavigateToHdScanNewAddresses -> {
+                // Will be added when hd widget is merged in
+            }
         }
     }
 
--- a/app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
+++ b/app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
@@ -15,6 +15,7 @@ package com.algorand.android.modules.accountdetail.accountstatusdetail.ui
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.viewModelScope
 import com.algorand.android.core.BaseViewModel
+import com.algorand.android.models.AccountCreation
 import com.algorand.android.models.AnnotatedString
 import com.algorand.android.models.ui.AccountAssetItemButtonState
 import com.algorand.android.modules.accountcore.ui.model.AccountDisplayName
@@ -25,9 +26,19 @@ import com.algorand.android.modules.accountdetail.accountstatusdetail.ui.Account
 import com.algorand.android.modules.accountdetail.accountstatusdetail.ui.AccountStatusDetailViewModel.ViewState
 import com.algorand.android.modules.accountdetail.accountstatusdetail.ui.decider.AccountStatusDetailPreviewDecider
 import com.algorand.android.modules.accounticon.ui.model.AccountIconDrawablePreview
-import com.algorand.android.modules.accounts.lite.domain.model.AccountLiteCacheStatus
-import com.algorand.android.modules.accounts.lite.domain.usecase.GetAccountLiteCacheFlow
+import com.algorand.android.utils.analytics.CreationType
+import com.algorand.android.utils.launchIO
+import com.algorand.wallet.account.core.domain.usecase.GetAccountDetailFlow
+import com.algorand.wallet.account.detail.domain.model.AccountType
 import com.algorand.wallet.account.detail.domain.model.AccountType.Companion.canSignTransaction
+import com.algorand.wallet.account.detail.domain.usecase.GetAccountRegistrationType
+import com.algorand.wallet.account.info.domain.usecase.GetAccountInformation
+import com.algorand.wallet.account.local.domain.model.LocalAccount
+import com.algorand.wallet.account.local.domain.usecase.GetHdEntropy
+import com.algorand.wallet.account.local.domain.usecase.GetHdKeyPrivateKey
+import com.algorand.wallet.account.local.domain.usecase.GetLocalAccount
+import com.algorand.wallet.encryption.domain.manager.AESPlatformManager
+import com.algorand.wallet.encryption.domain.utils.clearFromMemory
 import com.algorand.wallet.viewmodel.EventDelegate
 import com.algorand.wallet.viewmodel.EventViewModel
 import com.algorand.wallet.viewmodel.StateDelegate
@@ -37,6 +48,7 @@ import javax.inject.Inject
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.launch
 
+@Suppress("LongParameterList")
 @HiltViewModel
 class AccountStatusDetailViewModel @Inject constructor(
     savedStateHandle: SavedStateHandle,
@@ -46,7 +58,13 @@ class AccountStatusDetailViewModel @Inject constructor(
     private val getAccountIconDrawablePreview: GetAccountIconDrawablePreview,
     private val getAccountOriginalStateIconDrawablePreview: GetAccountOriginalStateIconDrawablePreview,
     private val accountStatusDetailPreviewDecider: AccountStatusDetailPreviewDecider,
-    private val getAccountLiteCacheFlow: GetAccountLiteCacheFlow
+    private val getAccountDetailFlow: GetAccountDetailFlow,
+    private val getAccountInformation: GetAccountInformation,
+    private val getAccountRegistrationType: GetAccountRegistrationType,
+    private val getLocalAccount: GetLocalAccount,
+    private val aesPlatformManager: AESPlatformManager,
+    private val getHdKeyPrivateKey: GetHdKeyPrivateKey,
+    private val getHdEntropy: GetHdEntropy
 ) : BaseViewModel(), StateViewModel<ViewState> by stateDelegate, EventViewModel<ViewEvent> by eventDelegate {
 
     private val navArgs = AccountStatusDetailBottomSheetArgs.fromSavedStateHandle(savedStateHandle)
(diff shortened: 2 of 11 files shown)


app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailBottomSheet.kt
```diff
@@ -82,6 +82,9 @@ class AccountStatusDetailBottomSheet : BaseBottomSheet(R.layout.bottom_sheet_acc
             is ViewEvent.NavigateToRekeyToLedgerAccount -> {
                 navToRekeyToLedgerAccountNavigation()
             }
+            is ViewEvent.NavigateToHdScanNewAddresses -> {
+                // Will be added when hd widget is merged in
```
yasin-ce, line 86: Please add a TODO here to not forget this.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -96,6 +136,7 @@ class AccountStatusDetailViewModel @Inject constructor(
                         accountTypeDrawablePreview = accountTypeDrawablePreview,
                         descriptionDetail = descriptionDetail,
                         accountTypeString = accountTypeString,
+                        isHdWallet = accountType == AccountType.HdKey,
```
yasin-ce, line 139: This parameter is not used. If it will be used please leave a comment.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -121,6 +162,39 @@ class AccountStatusDetailViewModel @Inject constructor(
         eventDelegate.sendEvent(viewModelScope, ViewEvent.NavigateToRekeyToLedgerAccount)
     }
 
+    fun navToHdScanNewAddresses() {
```
yasin-ce, line 165: This is not used. If it will be used please leave a comment.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -121,6 +162,39 @@ class AccountStatusDetailViewModel @Inject constructor(
...
+                                account.account,
+                                account.change,
+                                account.keyIndex,
+                                account.derivationType,
+                            ),
+                            creationType = CreationType.RECOVER
+                        )
+                        privateKey.clearFromMemory()
+                        entropy.clearFromMemory()
+                        eventDelegate.sendEvent(viewModelScope, ViewEvent.NavigateToHdScanNewAddresses(accountCreation))
```
yasin-ce, line 190: Instead of passing private data to the another screen we may generate this data at that screen to make it safer.

common-sdk/src/main/kotlin/com/algorand/wallet/account/custom/domain/usecase/GetHdSeedIdFromAddressUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License
+ */
+
+package com.algorand.wallet.account.custom.domain.usecase
+
+import com.algorand.wallet.account.local.domain.model.LocalAccount
+import com.algorand.wallet.account.local.domain.usecase.GetLocalAccount
+import javax.inject.Inject
+
+internal class GetHdSeedIdFromAddressUseCase @Inject constructor(
```
yasin-ce, line 19: Test class is missing.

yasin-ce (changes requested):

common-sdk/src/main/kotlin/com/algorand/wallet/account/detail/domain/model/AccountDetail.kt
```diff
@@ -13,11 +13,13 @@
 package com.algorand.wallet.account.detail.domain.model
 
 import com.algorand.wallet.account.custom.domain.model.CustomAccountInfo
+import com.algorand.wallet.account.custom.domain.model.CustomHdSeedInfo
 import com.algorand.wallet.account.detail.domain.model.AccountType.Companion.canSignTransaction
 
 data class AccountDetail(
     val address: String,
     val customAccountInfo: CustomAccountInfo?,
+    val customHdSeedInfo: CustomHdSeedInfo?,
```
mitsinsar, line 22: Instead of adding this as a new variable, please update CustomAccountInfo. Otherwise, we have to call 
`customAccountInfo.isBackedUp ?: customHdSeedInfo.isBackedUp ?: false` to get a single field. You can do something like below;

```
data class CustomAccountInfo(
    val customName: String?,
    val orderIndex: Int,
    val isBackedUp: Boolean,
    val detail: Detail
) {

    sealed interface Detail {
        data class Address(val address: String) : Detail
        data class HdSeed(val seedId: Int) : Detail
    }
}
```

I'm not sure about calling it a `detail`, but this'll give you an idea about how you should update it

common-sdk/src/main/kotlin/com/algorand/wallet/account/custom/domain/usecase/GetHdSeedIdFromAddressUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val getLocalAccount: GetLocalAccount
+) : GetHdSeedIdFromAddress {
+
+    override suspend fun invoke(address: String): Int {
```
mitsinsar, line 23: Completely wrong implementation. How can we trust the `id` it returns? you should either return PeraResult or just null, so that we can understand there is no seed id for given address

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -56,24 +74,46 @@ class AccountStatusDetailViewModel @Inject constructor(
...
     }
 
+    @Suppress("LongMethod")
     fun loadAccountStatusDetail() {
         stateDelegate.updateState { ViewState.Loading }
         viewModelScope.launch {
-            getAccountLiteCacheFlow().collectLatest { cacheStatus ->
-                val accountLite = (cacheStatus as? AccountLiteCacheStatus.Data)?.accountLites?.get(accountAddress)
-                if (accountLite?.cachedInfo == null) return@collectLatest
+            getAccountDetailFlow(accountAddress).collectLatest { accountDetail ->
```
mitsinsar, line 81: Don't use `getAccountDetailFlow`. We are getting rid of this function because it is not working properly for addresses with lots of assets. Keep using AccountLite please

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -56,24 +74,46 @@ class AccountStatusDetailViewModel @Inject constructor(
...
+                if (accountDetail == null) return@collectLatest
+                val accountType = accountDetail.accountType
 
-                val accountType = accountLite.cachedInfo.type
-
-                val authAccountAddress = accountLite.cachedInfo.rekeyAuthAddress
-                val hasAccountAuthority = accountType.canSignTransaction()
+                val accountInformation = getAccountInformation(accountAddress)
```
mitsinsar, line 85: Don't use `getAccountInformation`. AccountLite already has `rekeyAdminAddress`

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
```diff
@@ -56,24 +74,46 @@ class AccountStatusDetailViewModel @Inject constructor(
...
-                val accountOriginalTypeDisplayName = getAccountDisplayName(accountAddress)
+                val accountOriginalTypeDisplayName = if (accountType == AccountType.HdKey) {
+                    AccountDisplayName(
+                        accountAddress = accountAddress,
+                        primaryDisplayName = accountDetail.customHdSeedInfo?.entropyCustomName ?: accountAddress,
+                        secondaryDisplayName = accountDetail.customHdSeedInfo?.entropyCustomName ?: accountAddress
+                    )
+                } else {
+                    getAccountDisplayName(accountAddress)
+                }
```
mitsinsar, line 98: This happens if we don't update existing custom info model and add customHdSeedInfo... There are rules for setting display names. Never create AccountDisplayName manually. Address may have NFD. Also we never set accountAddress directly, it has to be shortened.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
mitsinsar: I agree on this. If destination requires `AccountCreation` then you should update it in a way that it doesn't require account creation, because it is not an account creation. This just pollutes the code.

mitsinsar (changes requested):

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: a point worth talking about tomorrow...we are still using `accountCreation` in onboarding flow for hd wallet, but we can try to remove it

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: i understand to switch back to `AccountLite`.  I think the problem for me is if we add hd seed custom data to `AccountLite` model, we're fetching it when 90% of time we don't use it.  Maybe we can use accountLite but only fetch hd seed custom data in this one viewmodel that needs it.

app/src/main/kotlin/com/algorand/android/modules/accountdetail/accountstatusdetail/ui/AccountStatusDetailViewModel.kt
michaeltchuang: there are two custom info names...one for account and one for root wallet name (hd case).  For hd, sometimes in UI we display custom account name and sometimes we display the custom wallet name.  I think maybe we need to create a `getWalletDisplayName` for hd case perhaps

michaeltchuang:
closing this and we can re-open another PR using the UI work in https://github.com/perawallet/pera-android/pull/307

PR #304, opened 2025-05-21
PERA-2237 :: Fix Algo25 with empty secret key issue

mitsinsar:
## Description

This was a known bug, and finally it will be resolved. Problem is, user is not able to sign transaction with Algo25 accounts. 
There are 2 ways to reproduce this issue; 
1. Undo Rekey from Account Detail
- Import an auth account with passphrase
- In the next screen, select a rekeyed account to import
- Go to AccountDetail
- Undo rekey

We had auth account (the one we added first), so we were able to sign transactions of the rekeyed account. But after this operation, user won’t be able to sign any transaction. This is expected, however app behaves like this rekeyed account is Algo25 account. Even if app displays Watch account or any other state, account is being updated as Algo25 with empty secret key. 
Since user added this account with address only (rekeyed account selection), we never had second account’s signing details.
2. Undo Rekey from other platforms
- Import a rekeyed account without using passphrase
- Undo rekey without using Pera
Both flows cause us to have Standard/Algo25 accounts without secret keys.

Importing should be done with prod version, not with 6.x.x
With 6.0.0, we won’t have this issue, but this empty secret key issue also effects migration. So, this ticket is to fix migration issue and also add validation for previously migrated apps (beta users)
We need to make sure migration works properly under these circumstances.
Users with Algo25 accounts without secret keys should see Watch Account after migration.
Users who have already migrated empty secret key Algo25 accounts should see watch account as well. 

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2237

## Checklist


michaeltchuang (approved):
it'd be nice if you could add a video of it working

also since migration has di/data/domain folder...maybe dragging whole thing into modules folder might make sense (align with our new architecture style)

common-sdk/src/main/kotlin/com/algorand/wallet/account/local/data/repository/DefaultAlgo25NoAuthRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+    override suspend fun updateInvalidAlgo25AccountsToNoAuth() {
+        val algo25Entities = algo25NoAuthDao.getAllAlgo25Entities()
+        val invalidAlgo25Addresses = algo25Entities.mapNotNull { entity ->
+            entity.algoAddress.takeIf { entity.isSecretKeyInvalid() }
+        }
+        val entities = invalidAlgo25Addresses.map { noAuthEntityMapper(it) }
+        algo25NoAuthDao.updateAlgo25AccountsToNoAuthAccounts(entities)
+    }
+
+    private fun Algo25Entity.isSecretKeyInvalid(): Boolean {
```
michaeltchuang, line 37: i was also thinking we could check the AlgoSDK if we can restore an account based off secret key and not just 0 check.  If this handles the defly rekey and undo rekey use case, then i guess that's the important part

PR #316, opened 2025-05-29
PERA-1582 :: Implement scan new addresses feature

mitsinsar:
## Description
- Passed entropy to RecoverRegisteredAccounts instead of account creation, since it only needs entropy
- Fixed registered account list divider padding and item height issue

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-1582

## Checklist
- [x] Have you tested your changes locally?

https://github.com/user-attachments/assets/cbcd724b-73ba-4c16-8b6c-936c6050e965

### DIFF
--- a/app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/RecoverWithPassphraseFragment.kt
+++ b/app/src/main/kotlin/com/algorand/android/modules/onboarding/recoverypassphrase/enterpassphrase/ui/RecoverWithPassphraseFragment.kt
@@ -166,10 +166,13 @@ class RecoverWithPassphraseFragment : DaggerBaseFragment(R.layout.fragment_recov
         }
     }
 
-    private val navToNameRegistrationScreenEventCollector: suspend (Event<AccountCreation>?) -> Unit =
-        {
-            it?.consume()?.run { navigateToSuccess(this) }
-        }
+    private val navToNameRegistrationScreenEventCollector: suspend (Event<AccountCreation>?) -> Unit = {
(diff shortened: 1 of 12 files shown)


michaeltchuang:
Rescan works fine for me on this PR...but i noticed when i go through normal import 24 word hd root wallet flow (then picking a derived address), it shows pin step and not address rename screen.  @yasin-ce what are you seeing or expecting?  This might be a separate bug

mitsinsar:
We never implemented address renaming for hd keys, but it is expected to see pin step when pin is not set and don't show again is not selected

michaeltchuang:
> We never implemented address renaming for hd keys, but it is expected to see pin step when pin is not set and don't show again is not selected

you are right on not having the custom name for hd (since there could be many addresses selected), but I think the hd import flow should match the algo25 one in not showing pin screen on import (and also show confetti afterwards).  @LoafPickleWW which direction do you wanna go with pin screen for hd import?

LoafPickleWW:
It should be import a universal wallet wallet/account > home screen with confetti 🎊 

We can add in a condition that if it's one account the rename screen appears, similar to algo25

CC @mitsinsar @michaeltchuang

yasin-ce (approved):

yasin-ce:
@mitsinsar could you create a new ticket for the feature requests below?

```
It should be import a universal wallet wallet/account > home screen with confetti 🎊
We can add in a condition that if it's one account the rename screen appears, similar to algo25
```

mitsinsar:
Ticket ref: https://algorandfoundation.atlassian.net/browse/PERA-2331

PR #336, opened 2025-06-18
PERA-2394 :: Clear seed table index when all items are deleted

yasin-ce:
# Pull Request Template

## Description
- Seed id was increasing from where it left even if you delete all seeds. Now we reset the seed id when all seeds are deleted.

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2394

## Checklist

## Additional Notes
- Add any additional notes or comments that may be helpful for reviewers.


common-sdk/src/main/kotlin/com/algorand/wallet/account/local/data/database/dao/HdSeedDao.kt
```diff
@@ -56,4 +56,7 @@ internal interface HdSeedDao {
 
     @Query("DELETE FROM hd_seeds")
     suspend fun clearAll()
+
+    @Query("DELETE FROM sqlite_sequence")
```
mitsinsar: Seems like this deletes all tables autoincrement value. I think it would be better to add `WHERE name = hd_seeds` to clear seed table only.

common-sdk/src/test/kotlin/com/algorand/wallet/account/local/data/repository/HdSeedRepositoryImplTest.kt
```diff
@@ -168,6 +167,15 @@ class HdSeedRepositoryImplTest {
...
     }
 
+    @Test
+    fun `EXPECT successful completion WHEN clearing primary key index`() = runTest {
+        coEvery { hdSeedDao.clearPrimaryKeyIndex() } returns Unit
+
+        val result = hdSeedDao.clearPrimaryKeyIndex()
+
+        assertEquals(Unit, result)
+    }
```
mitsinsar: This test is not testing the changes you added to `HdSeedRepositoryImpl`. You are calling dao function, not repository function. Since we need to test changes we made in repository, you need to call `sut` function to test those changes.

For the changes you made, you need to add 3 tests; 
1. `deleteHdSeed` -> clear needs to be triggered when table size is less than 1
2. `deleteHdSeed` -> no clear triggered when table size is not less than 1
3. `deleteAllHdSeeds` -> when this is called, we need to make sure `clearPrimaryKeyIndex` is being called.

mitsinsar (changes requested):

mitsinsar (approved):

PR #342, opened 2025-06-23
PERA-1927 Hide asset id in WC txn detail if the asset is Algo

mitsinsar:
Asset ID will be hidden when the asset is Algo

Screenshot 2025-06-23 at 12 30 51


wjbeau (approved):
utACK

PR #346, opened 2025-06-26
PERA-2236 Override wc txn request deeplink in webview client

mitsinsar:
## Description
When a web page redirects to another page, we check if we should let it navigate or handle it by ourself. In this case, we weren't handling the redirect url we were supposed to and since the url is a deep link, webview wasn't able to load it and was displaying an error. 

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2236

## Checklist

https://github.com/user-attachments/assets/8d72151e-eb47-4a8e-9ff3-ce6e3b597d62


app/build.gradle
```diff
@@ -67,6 +67,7 @@ android {
         buildConfigField "String", "GitHash", "\"${getGitHash()}\""
         buildConfigField "String", "APPLICATION_NAME", '"pera"'
         buildConfigField "String", "DEEPLINK_PREFIX", '"algorand://"'
+        buildConfigField "String", "PERA_WC_DEEPLINK_PREFIX", '"perawallet-wc://"'
```
wjbeau, line 70: Should we add `perawallet://` here too in case we ever want to deep link into a wallet page from the webview?

wjbeau (approved):
utACK

Just a question for future proofing, but looks good to me.

app/build.gradle
mitsinsar: If that's ok, I'd suggest to create an enhancement ticket for that and then define the scope. I can simply add it here but there may be cases that I didn't think of. I believe it would be safer to do this change after discussing in the meeting.

PR #354, opened 2025-07-09
PERA-2366 :: Implement account detail chart

mitsinsar:
## Related Issues
Closes https://algorandfoundation.atlassian.net/browse/PERA-2366

https://github.com/user-attachments/assets/0907df00-b8ca-4bae-b390-f22aa6dfb909


wjbeau (changes requested):
Looks good - same observation as on portfolio chart that we don't need the 1D option.  

Also I wonder if there's an opportunity to reuse a bit more of the code.  Feels like the data model between the two charts is identical, etc?  Just a thought, not a blocker.

wjbeau (approved):
Whoops, meant to approve, not request changes.  I'm assuming the 1D issue will get resolved with a rebase

yasin-ce:
Screenshot 2025-07-11 at 12 27 23

0 value state is presented differently on iOS and android. We can align with the product team about that. 
Android design is accepted as the correct one so no action needed here.

yasin-ce:
Info icons size and color is not matching with the design.

yasin-ce:
Screenshot 2025-07-11 at 13 20 21
It might not related to this ticket but action button backgrounds are not visible.

yasin-ce (approved):

PR #358, opened 2025-07-11
Merge release to dev

yasin-ce:
# Pull Request Template

## Description
- Merge release branch to dev branch


gradle/libs.versions.toml
```diff
@@ -10,7 +10,7 @@ android-minSdk = "28"
 android-namespace = "com.algorand.android"
 android-targetSdk = "34"
 android-versionCode = "124"
-android-versionName = "6.202511.2"
```
wjbeau, line 13: We should probably roll this version after release, just so any dev builds carry the new version number (i.e. 6.202513.0)

wjbeau (approved):
utACK - I typically have cherry-picked individual hot fixes onto the main/dev branch from release, so merges only go one way, but I guess that's probably just a personal preference.

gradle/libs.versions.toml
yasin-ce: Thanks for the feedback. I have few concerns about this approach. Tomorrow we can discuss and I can use the final decision for the next releases.

PR #361, opened 2025-07-15
PERA-2371 :: Navigation Analytics

mitsinsar:
# Pera Event Tracking Implementation Change

## Problem
Existing event tracking implementation has some issues like;
- It adds referral data (if exists) into every event, even for button click tracking
- It adds strong box using state (if exists) into every event, even for button click trackings
- We need to add "t_" prefix to event names when the selected node is testnet. This operation is being done by `BaseEventTracker`. So fragment event trackers need to extend `BaseEventTracker`. However, new developers may not know that and use `PeraEventTracker` directly, which causes missing "t_" prefix for testnet events.
- `PeraEventTracker` functions are suspended, which requires `coroutineScope`. We usually use ViewModelScope for these operations. However, if we trigger event logging after calling some function that causes VM to die (navigating to another screen), coroutineScope dies and it causes logging to fail. Suspend functions also causes us to create redundant functions in ViewModels (can be seen in the screenshot)

Screenshot 2025-07-15 at 15 47 26

## Solution

- I created `PeraAnalyticsEventTracker` which normalizes name based on selected node and uses `PeraAnalyticsRepository` to log events.

- Implementation of `PeraAnalyticsRepository` is in `app` module, not `common-sdk` because it uses Firebase which requires Bundle for payload, and bundle is android specific. I try to keep `common-sdk` as shareable as possible. 

- `PeraAnalyticsEventTracker` has its own scope, which prevents dead coroutine scopes and also we can log events from fragment directly without adding redundant functions to ViewModels (by using delegation) 

Screenshot 2025-07-15 at 15 47 42

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2371

## Checklist
- [ ] Have you reviewed the code for any potential issues?


wjbeau (approved):
utACK

yasin-ce (approved):

PR #378, opened 2025-08-04
PERA-2609 Implement swap domain and date modules

mitsinsar:
## Description
PR targets to `PERA-2609` which is ticket for swap widget.  Created sub-branches to create smaller(!) PRs. UI will be built on top of this.

## Related Issues
- First part of https://algorandfoundation.atlassian.net/browse/PERA-2609

## Checklist

# UI will be improved with next PR
https://github.com/user-attachments/assets/0f2ca3dc-e24c-4681-ada8-614a7fe366f5


common-sdk/src/main/kotlin/com/algorand/wallet/swap/data/service/SwapApiService.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @POST("v1/dex-swap/prepare-transactions/")
+    suspend fun getQuoteTransactions(
+        @Body requestBody: CreateSwapQuoteTransactionsRequestBody
+    ): CreateSwapQuoteTransactionsResponse
+
+    @PATCH("v1/dex-swap/quotes/{quote_id}/")
+    suspend fun updateSwapQuoteException(
+        @Path("quote_id") quoteId: Long,
+        @Body swapQuoteExceptionRequestBody: SwapQuoteExceptionRequestBody
```
wjbeau, line 53: I suspect these will change to v2 soon

common-sdk/src/main/kotlin/com/algorand/wallet/swap/domain/model/SwapQuoteProvider.kt
```diff
@@ -0,0 +1,23 @@
...
+
+enum class SwapQuoteProvider {
+    TINYMAN,
+    TINYMAN_V2,
+    TINYMAN_SWAP_ROUTER,
+    VESTIGE_V3,
+    VESTIGE_V4,
+    FOLKS,
+    DEFLEX
+}
```
wjbeau, line 23: I think Samet has come up with a set of rules to validate the swaps on the device without having to have provider specific validation code.  That would mean we can easily add future providers on the backend without requiring client side changes.  However, if we have this enum (and the peer ones) it removes that ability.  It seems like this is only used to validate the API responses correct?  Would it make sense to just use a String?  We will obviously need some code somewhere to show the right display name and icon, but maybe we change the server response payload to include those or something?

wjbeau (approved):
Looks good - just a couple minor thoughts

PR #396, opened 2025-09-03
PERA-2701 :: Create new asset detail screen

mitsinsar:
## Description
There are still missing things like; fav button, notification button, swap history
There is no entry point to this screen yet, so it is safe to merge with missing implementations.

## Related Issues
- https://algorandfoundation.atlassian.net/browse/PERA-2701

https://github.com/user-attachments/assets/f7998328-102d-498b-a6c8-61a804d9bed2


wjbeau (approved):
You mentioned in the comment that swap history is missing, but just to clarify that the bottom half of this screen is actually a revised transaction history, not just swaps.

yasin-ce (approved):

PR #426, opened 2025-10-07
PERA-2831 Create credentials module

mitsinsar:
### Module Structure
The core logic for passkey creation and assertion has been encapsulated in a new `credentials` module. However, passkeys screen is still in `app` module because our design components are in `app` module.

### Registration Flow
- **User Handle:** User's username is used as userHandle during the key pair generation process
- **Credential ID:** is derived from generated key pair
- **Duplicate prevention:** To check if credential already exists, username, url and seed id combination are used.

### Authentication Flow
- During authentication, `userId` is used for `userHandle` field in `AuthenticatorAssertionResponse`

### Notes
You may see some classes that uses `JSONObject` and hardcoded Json keys. These classes have been copied directly from `androidx.credentials` library. This was done intentionally to handle specific FIDO2 standards without adding new dependency or mappers for a few classes.

## Related Issues
Closes https://algorandfoundation.atlassian.net/browse/PERA-2831


wjbeau (approved):
Looks good - we might want to review some of the error message strings if these are shown to users since the user won't really have much of an idea what they mean.

PR #431, opened 2025-10-08
PERA-2472 :: Fix assert sorting issues

mitsinsar:
## Description
While calculating `total_usd_value` as sorting value, asset decimal wasn't included into calculation. That's why sorting wasn't working properly. Android SQLite doesn't have functions like pow, and it doesn't let us load custom functions. To solve the issue, I used scientific notation.

Also added 2 tie breakers;
- If usd values are the same, items will be sorted alphabetically
- If usd values and names are the same, items will be sorted by asset id.

## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2472


common-sdk/src/main/kotlin/com/algorand/wallet/asset/data/database/dao/PaginatedAssetCollectibleDao.kt
```diff
@@ -26,7 +26,7 @@ internal interface PaginatedAssetCollectibleDao {
         holding.algo_address AS algo_address,
         holding.amount AS amount,
         holding.asset_status AS asset_status,
-        (holding.amount * IFNULL(asset.usd_value, 0)) AS total_usd_value,
+        (CAST(holding.amount AS REAL) / CAST('1e' || IFNULL(asset.decimals, 0) AS REAL)) * IFNULL(asset.usd_value, 0) AS total_usd_value,
```
wjbeau: Looks okay but I'm guessing if you have 18 decimals that won't fit in a REAL?  Should we maybe be using DECIMALs?

wjbeau (approved):

common-sdk/src/main/kotlin/com/algorand/wallet/asset/data/database/dao/PaginatedAssetCollectibleDao.kt
mitsinsar: Nice catch, thank you. I didn't consider this case since total_usd_value is used for sorting only, but we may use it to display the value in the future.

PR #427, opened 2025-10-07
PERA-2831 Implement passkeys data and domain packages

mitsinsar:
## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-2831

Details can be found in parent PR: https://github.com/perawallet/pera-android/pull/426#issue-3493149529

PR #436, opened 2025-10-10
PERA-2931 Use address instaed of seed id for passkeys

mitsinsar:
Created separate PR to make it easier to review/discuss about changes

### Resolved comments;

[Seed Id - Address](https://github.com/perawallet/pera-android/pull/427#discussion_r2412949029)
[Passkey/KeyPair relation](https://github.com/perawallet/pera-android/pull/427#discussion_r2412964074)

### Discussions;

[Kotlinx Cbor](https://github.com/perawallet/pera-android/pull/427#discussion_r2412991651)

Using Kotlinx Cbor serializer makes sense, but to be able to do that, I have to create AttestationObjectResult types, and create custom serializers for those types. I couldn't decide if it's worth doing that for initial version or not. If you prefer creating types and serializers, I can do it. 

## Related Issues
Closes https://algorandfoundation.atlassian.net/browse/PERA-2831


credentials/src/main/java/com/algorand/android/credentials/passkeys/domain/Bip39SignManager.kt
```diff
@@ -14,8 +14,8 @@ package com.algorand.android.credentials.passkeys.domain
 
 import java.security.KeyPair
 
-internal interface PasskeyManager {
-    suspend fun signPasskey(seedId: Int, origin: String, userHandle: String, payload: ByteArray): ByteArray
-    suspend fun derivePasskey(seedId: Int, origin: String, userHandle: String): KeyPair
+internal interface Bip39SignManager {
```
wjbeau, line 17: I think this is a P256SignManager no?

credentials/src/main/java/com/algorand/android/credentials/passkeys/domain/model/AddPasskeyArgs.kt
```diff
@@ -15,7 +15,7 @@ package com.algorand.android.credentials.passkeys.domain.model
 data class AddPasskeyArgs(
     val siteUrl: String,
     val siteName: String,
-    val seedId: Int,
+    val bip39Address: String,
```
wjbeau: To be super correct, I think it's technically a BIP44 address

wjbeau (approved):
Couple of naming nits but fine to do those in separate PR - not critical.  Also I agree the CBOR stuff isn't worth it if it requires a bunch of extra stuff.

credentials/src/main/java/com/algorand/android/credentials/passkeys/domain/Bip39SignManager.kt
mitsinsar: I named this `Bip39SignManager` because it is responsible of getting seed details and uses them to sign.
Calling it `P256SignManager` may cause us to give more responsibility in the future. I just wanted to limit this class' responsibility to "get seed details" and "sign". If this doesn't make sense, I can update as you suggested.

PR #496, opened 2025-12-15
PERA-2619 :: Implement swap transaction validation

mitsinsar:
## Description
- Implemented 4 transaction validation rule for swap transactions as below.
- Created feature toggle to enable/disable these validations

### Close to
- Checks if there is any transaction that closes a local address.
- Checks `closeToAddress` and `assetCloseToAddress` fields and compares sender address with local addresses.

### Signer
- Checks if there is any transaction where the sender matches any of the other local addresses (excluding the current swap account).
- Checks the `senderAddress` field of the signed/unsigned transactions and compares it with the list of local addresses.

### Rekey
- Checks if there is any transaction that rekeys any of the local addresses.
- Checks `rekeyAddress` field and compares sender address with local addresses.

### Sending Amount
- Checks if the total outgoing amount from the user's account exceeds the maximum expected amount .

## Related Issues
Closes https://algorandfoundation.atlassian.net/browse/PERA-2619


common-sdk/src/main/kotlin/com/algorand/wallet/swap/domain/validation/rules/SwapReceivedAmountValidationRule.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.algorand.wallet.algosdk.transaction.model.RawTransaction
+import com.algorand.wallet.algosdk.transaction.model.RawTransactionType
+import com.algorand.wallet.swap.domain.model.SwapQuoteV2
+import com.algorand.wallet.swap.domain.usecase.GetParsedSwapTransactions
+import com.algorand.wallet.swap.domain.validation.SwapTransactionValidationRule
+import com.algorand.wallet.swap.domain.validation.model.SwapTransactionValidationData
+import java.math.BigInteger
+import javax.inject.Inject
+
+internal class SwapReceivedAmountValidationRule @Inject constructor(
```
wjbeau, line 30: Should we maybe add a comment here for future us to remind us that this isn't fully implemented?

common-sdk/src/main/kotlin/com/algorand/wallet/swap/domain/validation/rules/SwapRekeyedTransactionValidationRule.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun invoke(data: SwapTransactionValidationData): Boolean {
+        val transactions = getParsedSwapTransactions(data.signedTransactions, data.unsignedTransactions)
+        return transactions.none { it.isRekeyingLocalAddress(data.localAddresses) }
+    }
+
+    private fun RawTransaction.isRekeyingLocalAddress(localAddresses: List<String>): Boolean {
+        val hasRekeyAddress = rekeyAddress != null
+        val isSenderLocalAddress = localAddresses.any { it == senderAddress?.decodedAddress }
+        return hasRekeyAddress && isSenderLocalAddress
```
wjbeau: Not a big deal but presumably we already have this logic implemented for the rekey warning in transaction signing.  Maybe we can reuse so we're only maintaining it once?

wjbeau (approved):

PR #501, opened 2026-01-12
PERA-3158 Log passkey click events

mitsinsar:
## Related Issues
- Closes https://algorandfoundation.atlassian.net/browse/PERA-3158

Screenshot 2026-01-12 at 15 34 50


wjbeau (approved):
Are we able to log the usage of passkeys on Android?  If so that might be really useful (.e.g when a passkey is created and/or used to sign in). Just to get some sense of how much the feature is actually used (vs how many people look at the settings screen which is a much rougher proxy for usage)

